PR #16, opened 2023-12-15
Copy Kokoro shared install lib

sergiitk:
The first step of moving the source of truth of the Kokoro shared install lib to the new repo.

Source: https://github.com/grpc/grpc/blob/4f7ead2c7b7b82ad80799bbfa482184991dcd346/tools/internal_ci/linux/grpc_xds_k8s_install_test_driver.sh
New name: `.kokoro/psm_interop_kokoro_lib.sh`

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -1,0 +1,524 @@
+#!/usr/bin/env bash
+# Copyright 2020 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -0,0 +1,524 @@
+#!/usr/bin/env bash
+# Copyright 2020 gRPC authors.
```
XuanWang-Amos, line 2: [Minor] Should be 2023 now

XuanWang-Amos (approved):

.kokoro/psm_interop_kokoro_lib.sh
sergiitk: That's the year the code was created, and this is 1:1 copy. We don't update the copyright in this case.

PR #20, opened 2024-01-13
Paramatize request/response payload size

stanley-cheung:
Currently, the xds interop client is sending an empty payload for the `UnaryCall` rpc. Since testing CSM Observability will need a non-trivial payload (e.g. to verify the metrics about the number of bytes being sent), this PR will try to add 2 parameters `request_payload_size` and `response_payload_size` to the client image to control the size of the payload to be sent. Each language will need to implement this parameter in their respective client image. The one for c++ is [here](https://github.com/grpc/grpc/pull/35545).

Note that we only need to set these 2 parameters (both `request` and `resposne` payload size) on the _client deployment_ because the client will in turn tell the server how much to send back.

The good 'ol interop client already has these: https://github.com/grpc/grpc/blob/master/test/cpp/interop/interop_client.cc#L249-L251

### DIFF
--- a/framework/test_app/runners/k8s/k8s_xds_client_runner.py
+++ b/framework/test_app/runners/k8s/k8s_xds_client_runner.py
@@ -156,6 +156,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
             server_target=server_target,
             rpc=rpc,
             qps=qps,
+            request_payload_size=271828,
+            response_payload_size=314159,
             metadata=metadata,
             secure_mode=secure_mode,
             config_mesh=config_mesh,
--- a/framework/test_app/runners/k8s/k8s_xds_client_runner.py
+++ b/framework/test_app/runners/k8s/k8s_xds_client_runner.py
@@ -107,6 +107,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
         print_response=False,
         log_to_stdout: bool = False,
         enable_csm_observability: bool = False,
+        request_payload_size: int = 0,
+        response_payload_size: int = 0,
     ) -> XdsTestClient:
         logger.info(
             (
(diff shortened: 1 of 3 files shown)


kubernetes-manifests/client.deployment.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           - "--qps=${qps}"
           - "--rpc=${rpc}"
           - "--metadata=${metadata}"
+          - "--request_payload_size=${request_payload_size}"
```
sergiitk, line 44: These should be behind an if , similar to `% if enable_csm_observability:`

framework/test_app/runners/k8s/k8s_xds_client_runner.py
```diff
@@ -156,6 +156,8 @@ def run(  # pylint: disable=arguments-differ
             server_target=server_target,
             rpc=rpc,
             qps=qps,
+            request_payload_size=271828,
```
sergiitk, line 159: These should be arguments of the `run` method, and set to 0 by default.

sergiitk (changes requested):

kubernetes-manifests/client.deployment.yaml
stanley-cheung: Fixed.

framework/test_app/runners/k8s/k8s_xds_client_runner.py
stanley-cheung: Fixed

stanley-cheung: Fixed

sergiitk (approved):
All good - had a comment about the constants, but #23 takes care of it

framework/test_app/runners/k8s/k8s_xds_client_runner.py
sergiitk, line 1: Please make sure to rebase after #24 is merged to the main, and run the CSM test,

sergiitk: - [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/89a5b55f-3ce5-4141-a9c0-0483aa6a1eb3)

stanley-cheung: This job has passed!

sergiitk: You could press "resolve conversation" - that would've triggered the auto-merge :)

PR #23, opened 2024-01-20
Add assertions to CSM Observability Test

stanley-cheung:
This PR adds some assertions to test whether the metrics generated by the CSM Observability Test match what we expect.

We use the `list_time_series` cloud monitoring API to query metrics we just generated by the test run, and then verify whether the metrics data (label keys and values, and data points) is what we expected.

### DIFF
--- a/framework/test_app/runners/k8s/k8s_xds_client_runner.py
+++ b/framework/test_app/runners/k8s/k8s_xds_client_runner.py
@@ -94,7 +94,7 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
             # permission to use GCP service account identity.
             self.gcp_iam = gcp.iam.IamV1(gcp_api_manager, gcp_project)
 
-    def run(  # pylint: disable=arguments-differ
+    def run(  # pylint: disable=arguments-differ,too-many-locals
         self,
         *,
         server_target,
@@ -109,6 +109,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
         enable_csm_observability: bool = False,
         request_payload_size: int = 0,
         response_payload_size: int = 0,
+        csm_workload_name: str = "",
+        csm_canonical_service_name: str = "",
     ) -> XdsTestClient:
         logger.info(
             (
@@ -166,6 +168,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
             generate_mesh_id=generate_mesh_id,
             print_response=print_response,
             enable_csm_observability=enable_csm_observability,
+            csm_workload_name=csm_workload_name,
+            csm_canonical_service_name=csm_canonical_service_name,
         )
 
         # Create a PodMonitoring resource if CSM Observability is enabled
--- a/framework/test_app/runners/k8s/k8s_xds_client_runner.py
+++ b/framework/test_app/runners/k8s/k8s_xds_client_runner.py
@@ -34,6 +34,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
     debug_use_port_forwarding: bool
     td_bootstrap_image: str
     network: str
+    csm_workload_name: str
+    csm_canonical_service_name: str
 
     # Optional fields.
     service_account_name: Optional[str] = None
@@ -62,6 +64,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
         namespace_template: Optional[str] = None,
         debug_use_port_forwarding: bool = False,
         enable_workload_identity: bool = True,
+        csm_workload_name: str = "",
+        csm_canonical_service_name: str = "",
     ):
         super().__init__(
             k8s_namespace,
@@ -79,6 +83,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
         self.deployment_template = deployment_template
         self.enable_workload_identity = enable_workload_identity
         self.debug_use_port_forwarding = debug_use_port_forwarding
+        self.csm_workload_name = csm_workload_name
+        self.csm_canonical_service_name = csm_canonical_service_name
 
         # Used by the TD bootstrap generator.
         self.td_bootstrap_image = td_bootstrap_image
@@ -109,8 +115,6 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
         enable_csm_observability: bool = False,
         request_payload_size: int = 0,
         response_payload_size: int = 0,
-        csm_workload_name: str = "",
-        csm_canonical_service_name: str = "",
     ) -> XdsTestClient:
         logger.info(
             (
@@ -168,8 +172,8 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
             generate_mesh_id=generate_mesh_id,
             print_response=print_response,
             enable_csm_observability=enable_csm_observability,
-            csm_workload_name=csm_workload_name,
-            csm_canonical_service_name=csm_canonical_service_name,
+            csm_workload_name=self.csm_workload_name,
+            csm_canonical_service_name=self.csm_canonical_service_name,
         )
 
         # Create a PodMonitoring resource if CSM Observability is enabled
--- a/framework/test_app/runners/k8s/k8s_xds_client_runner.py
+++ b/framework/test_app/runners/k8s/k8s_xds_client_runner.py
@@ -100,7 +100,7 @@ class KubernetesClientRunner(k8s_base_runner.KubernetesBaseRunner):
             # permission to use GCP service account identity.
             self.gcp_iam = gcp.iam.IamV1(gcp_api_manager, gcp_project)
 
-    def run(  # pylint: disable=arguments-differ,too-many-locals
+    def run(  # pylint: disable=arguments-differ
         self,
         *,
         server_target,
--- a/kubernetes-manifests/client.deployment.yaml
+++ b/kubernetes-manifests/client.deployment.yaml
@@ -66,6 +66,22 @@ spec:
             value: "true"
           - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
             value: "true"
+          % if enable_csm_observability:
+          - name: CSM_WORKLOAD_NAME
+            value: ${csm_workload_name}
+          - name: CSM_CANONICAL_SERVICE_NAME
+            value: ${csm_canonical_service_name}
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: NAMESPACE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: OTEL_RESOURCE_ATTRIBUTES
+            value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
+          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/kubernetes-manifests/client.deployment.yaml
+++ b/kubernetes-manifests/client.deployment.yaml
@@ -74,6 +74,7 @@ spec:
           - name: CSM_CANONICAL_SERVICE_NAME
             value: ${csm_canonical_service_name}
           % endif
+          % if enable_csm_observability:
           - name: POD_NAME
             valueFrom:
               fieldRef:
@@ -84,6 +85,7 @@ spec:
                 fieldPath: metadata.namespace
           - name: OTEL_RESOURCE_ATTRIBUTES
             value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
+          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/kubernetes-manifests/client.deployment.yaml
+++ b/kubernetes-manifests/client.deployment.yaml
@@ -66,11 +66,14 @@ spec:
             value: "true"
           - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
             value: "true"
-          % if enable_csm_observability:
+          % if csm_workload_name:
           - name: CSM_WORKLOAD_NAME
             value: ${csm_workload_name}
+          % endif
+          % if csm_canonical_service_name:
           - name: CSM_CANONICAL_SERVICE_NAME
             value: ${csm_canonical_service_name}
+          % endif
           - name: POD_NAME
             valueFrom:
               fieldRef:
@@ -81,7 +84,6 @@ spec:
                 fieldPath: metadata.namespace
           - name: OTEL_RESOURCE_ATTRIBUTES
             value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
-          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/kubernetes-manifests/server.deployment.yaml
+++ b/kubernetes-manifests/server.deployment.yaml
@@ -60,6 +60,7 @@ spec:
           - name: CSM_CANONICAL_SERVICE_NAME
             value: ${csm_canonical_service_name}
           % endif
+          % if enable_csm_observability:
           - name: POD_NAME
             valueFrom:
               fieldRef:
@@ -70,6 +71,7 @@ spec:
                 fieldPath: metadata.namespace
           - name: OTEL_RESOURCE_ATTRIBUTES
             value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
+          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/kubernetes-manifests/server.deployment.yaml
+++ b/kubernetes-manifests/server.deployment.yaml
@@ -52,6 +52,22 @@ spec:
             value: "true"
           - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
             value: "true"
+          % if enable_csm_observability:
+          - name: CSM_WORKLOAD_NAME
+            value: ${csm_workload_name}
+          - name: CSM_CANONICAL_SERVICE_NAME
+            value: ${csm_canonical_service_name}
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: NAMESPACE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: OTEL_RESOURCE_ATTRIBUTES
+            value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
+          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/kubernetes-manifests/server.deployment.yaml
+++ b/kubernetes-manifests/server.deployment.yaml
@@ -52,11 +52,14 @@ spec:
             value: "true"
           - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
             value: "true"
-          % if enable_csm_observability:
+          % if csm_workload_name:
           - name: CSM_WORKLOAD_NAME
             value: ${csm_workload_name}
+          % endif
+          % if csm_canonical_service_name:
           - name: CSM_CANONICAL_SERVICE_NAME
             value: ${csm_canonical_service_name}
+          % endif
           - name: POD_NAME
             valueFrom:
               fieldRef:
@@ -67,7 +70,6 @@ spec:
                 fieldPath: metadata.namespace
           - name: OTEL_RESOURCE_ATTRIBUTES
             value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
-          % endif
         volumeMounts:
           - mountPath: /tmp/grpc-xds/
             name: grpc-td-conf
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -96,12 +96,19 @@ GammaServerRunner = gamma_server_runner.GammaServerRunner
 KubernetesClientRunner = k8s_xds_client_runner.KubernetesClientRunner
 
 
+# This class should represent one TimeSeries object from
+# monitoring_v3.ListTimeSeriesResponse.
 @dataclasses.dataclass(eq=False)
 class MetricTimeSeries:
+    # the metric name
     name: str
+    # each time series has a set of metric labels
     metric_labels: dict[str, str]
-    resource_labels: dict[str, str]
+    # each time series has a monitored resource
     resource_type: str
+    # each time series has a set of monitored resource labels
+    resource_labels: dict[str, str]
+    # each time series has a set of data points
     points: list[monitoring_v3.types.Point]
 
     @classmethod
@@ -113,8 +120,8 @@ class MetricTimeSeries:
         return cls(
             name=name,
             metric_labels=dict(response.metric.labels),
-            resource_labels=dict(response.resource.labels),
             resource_type=response.resource.type,
+            resource_labels=dict(response.resource.labels),
             points=list(response.points),
         )
 
@@ -139,12 +146,16 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
         super().setUpClass()
         cls.metric_client = cls.gcp_api_manager.monitoring_metric_service("v3")
 
+    # These parameters are more pertaining to the test itself, not to
+    # each run().
     def initKubernetesClientRunner(self, **kwargs) -> KubernetesClientRunner:
         return super().initKubernetesClientRunner(
             csm_workload_name=CSM_WORKLOAD_NAME_CLIENT,
             csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_CLIENT,
         )
 
+    # These parameters are more pertaining to the test itself, not to
+    # each run().
     def initKubernetesServerRunner(self, **kwargs) -> GammaServerRunner:
         return super().initKubernetesServerRunner(
             csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
@@ -249,7 +260,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
         with self.subTest("3_test_server_received_rpcs_from_test_client"):
             self.assertSuccessfulRpcs(test_client)
 
-        with self.subTest("4_query_monitoring_metric_client"):
+        with self.subTest("4_query_cloud_monitoring_metrics"):
             end_secs = int(time.time())
             interval = monitoring_v3.TimeInterval(
                 {
@@ -273,7 +284,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the correct set of metric keys and
         # values
-        with self.subTest("6_check_metrics_labels"):
+        with self.subTest("6_check_metrics_labels_histogram_client"):
             expected_metric_labels = {
                 "csm_mesh_id": unittest.mock.ANY,
                 "csm_remote_workload_canonical_service": CSM_CANONICAL_SERVICE_NAME_SERVER,
@@ -301,7 +312,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the correct set of metric keys and
         # values
-        with self.subTest("7_check_metrics_labels"):
+        with self.subTest("7_check_metrics_labels_histogram_server"):
             expected_metric_labels = {
                 "csm_mesh_id": unittest.mock.ANY,
                 "csm_remote_workload_canonical_service": CSM_CANONICAL_SERVICE_NAME_CLIENT,
@@ -326,7 +337,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the correct set of metric keys and
         # values
-        with self.subTest("8_check_metrics_labels"):
+        with self.subTest("8_check_metrics_labels_counter_client"):
             expected_metric_labels = {
                 "grpc_method": GRPC_METHOD_NAME,
                 "grpc_target": unittest.mock.ANY,
@@ -342,7 +353,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the correct set of metric keys and
         # values
-        with self.subTest("9_check_metrics_labels"):
+        with self.subTest("9_check_metrics_labels_counter_server"):
             expected_metric_labels = {
                 "grpc_method": GRPC_METHOD_NAME,
                 "otel_scope_name": unittest.mock.ANY,
@@ -357,7 +368,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the right set of monitored resource
         # label keys and values
-        with self.subTest("10_check_client_resource_labels"):
+        with self.subTest("10_check_client_resource_labels_client"):
             # all metrics should have the same set of monitored resource labels
             # keys, which come from the GMP job
             expected_resource_labels = {
@@ -381,7 +392,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
         # Testing whether each metric has the right set of monitored resource
         # label keys and values
-        with self.subTest("11_check_server_resource_labels"):
+        with self.subTest("11_check_server_resource_labels_server"):
             # all metrics should have the same set of monitored resource labels
             # keys, which come from the GMP job
             expected_resource_labels = {
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -125,10 +125,11 @@ class MetricTimeSeries:
             points=list(response.points),
         )
 
-    def pretty_print(self):
+    def pretty_print(self) -> str:
         metric = dataclasses.asdict(self)
+        # too much noise to print all data points from a time series
         metric.pop("points")
-        logger.info(yaml.dump(metric, sort_keys=True))
+        return yaml.dump(metric, sort_keys=True)
 
 
 class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
@@ -218,7 +219,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
             metric_time_series = MetricTimeSeries.from_response(
                 metric, time_series[0]
             )
-            metric_time_series.pretty_print()
+            logger.info(metric_time_series.pretty_print())
             results[metric] = metric_time_series
         return results
 
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -25,6 +25,8 @@ import yaml
 from framework import xds_gamma_testcase
 from framework import xds_k8s_testcase
 from framework.helpers import skips
+from framework.test_app.runners.k8s import gamma_server_runner
+from framework.test_app.runners.k8s import k8s_xds_client_runner
 
 logger = logging.getLogger(__name__)
 flags.adopt_module_key_flags(xds_k8s_testcase)
@@ -90,6 +92,9 @@ CLIENT_METRICS = HISTOGRAM_CLIENT_METRICS + COUNTER_CLIENT_METRICS
 SERVER_METRICS = HISTOGRAM_SERVER_METRICS + COUNTER_SERVER_METRICS
 ALL_METRICS = HISTOGRAM_METRICS + COUNTER_METRICS
 
+GammaServerRunner = gamma_server_runner.GammaServerRunner
+KubernetesClientRunner = k8s_xds_client_runner.KubernetesClientRunner
+
 
 @dataclasses.dataclass(eq=False)
 class MetricTimeSeries:
@@ -133,6 +138,18 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.metric_client = cls.gcp_api_manager.monitoring_metric_service("v3")
+
+    def initKubernetesClientRunner(self) -> KubernetesClientRunner:
+        return super().initKubernetesClientRunner(
+            csm_workload_name=CSM_WORKLOAD_NAME_CLIENT,
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_CLIENT,
+        )
+
+    def initKubernetesServerRunner(self) -> GammaServerRunner:
+        return super().initKubernetesServerRunner(
+            csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
+        )
 
     BuildQueryFn = Callable[[str], str]
 
@@ -217,8 +234,6 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
             start_secs = int(time.time())
             test_server: _XdsTestServer = self.startTestServers(
                 enable_csm_observability=True,
-                csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
-                csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
             )[0]
 
         with self.subTest("2_start_test_client"):
@@ -227,8 +242,6 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
                 enable_csm_observability=True,
                 request_payload_size=REQUEST_PAYLOAD_SIZE,
                 response_payload_size=RESPONSE_PAYLOAD_SIZE,
-                csm_workload_name=CSM_WORKLOAD_NAME_CLIENT,
-                csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_CLIENT,
             )
             logger.info("Letting test client run for %d seconds", TEST_RUN_SECS)
             time.sleep(TEST_RUN_SECS)
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -13,10 +13,10 @@
 # limitations under the License.
 import logging
 import time
+import unittest
 
 from absl import flags
 from absl.testing import absltest
-from google.api_core import exceptions as gapi_errors
 from google.cloud import monitoring_v3
 
 from framework import xds_gamma_testcase
@@ -35,7 +35,6 @@ _Lang = skips.Lang
 TEST_RUN_SECS = 90
 REQUEST_PAYLOAD_SIZE = 271828
 RESPONSE_PAYLOAD_SIZE = 314159
-VALUE_NOT_CHECKED = object()  # Sentinel value
 GRPC_METHOD_NAME = "grpc.testing.TestService/UnaryCall"
 CSM_WORKLOAD_NAME_SERVER = "csm_workload_name_from_server"
 CSM_WORKLOAD_NAME_CLIENT = "csm_workload_name_from_client"
@@ -70,22 +69,18 @@ METRIC_SERVER_CALL_DURATION = (
 METRIC_SERVER_CALL_STARTED = (
     f"{PROMETHEUS_HOST}/grpc_server_call_started_total/counter"
 )
-HISTOGRAM_CLIENT_METRICS = [
+HISTOGRAM_CLIENT_METRICS = (
     METRIC_CLIENT_ATTEMPT_SENT,
     METRIC_CLIENT_ATTEMPT_RCVD,
     METRIC_CLIENT_ATTEMPT_DURATION,
-]
-HISTOGRAM_SERVER_METRICS = [
+)
+HISTOGRAM_SERVER_METRICS = (
     METRIC_SERVER_CALL_DURATION,
     METRIC_SERVER_CALL_RCVD,
     METRIC_SERVER_CALL_SENT,
-]
-COUNTER_CLIENT_METRICS = [
-    METRIC_CLIENT_ATTEMPT_STARTED,
-]
-COUNTER_SERVER_METRICS = [
-    METRIC_SERVER_CALL_STARTED,
-]
+)
+COUNTER_CLIENT_METRICS = (METRIC_CLIENT_ATTEMPT_STARTED,)
+COUNTER_SERVER_METRICS = (METRIC_SERVER_CALL_STARTED,)
 HISTOGRAM_METRICS = HISTOGRAM_CLIENT_METRICS + HISTOGRAM_SERVER_METRICS
 COUNTER_METRICS = COUNTER_CLIENT_METRICS + COUNTER_SERVER_METRICS
 CLIENT_METRICS = HISTOGRAM_CLIENT_METRICS + COUNTER_CLIENT_METRICS
@@ -95,8 +90,6 @@ ALL_METRICS = HISTOGRAM_METRICS + COUNTER_METRICS
 
 class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
     metric_client: monitoring_v3.MetricServiceClient
-    test_server: _XdsTestServer
-    test_client: _XdsTestClient
 
     @staticmethod
     def is_supported(config: skips.TestConfig) -> bool:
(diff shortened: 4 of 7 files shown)


stanley-cheung:
Early indication is that @yashykt and I will need to debug some of the issues around some labels being missing.

stanley-cheung:
> Early indication is that @yashykt and I will need to debug some of the issues around some labels being missing. But the structure of the PR should be ready to be reviewed.

This has been resolved. Now all the expected CSM metric labels are showing up and the tests are now asserting against the values.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                    "csm_remote_workload_namespace_name",
+                    "csm_remote_workload_project_id",
+                    "csm_remote_workload_type",
+                    "csm_service_name",
+                    "csm_service_namespace_name",
+                    "csm_workload_canonical_service",
+                    "grpc_method",
+                    "grpc_status",
+                    "grpc_target",
+                    "otel_scope_name",
```
yashykt, line 253: i don't care about checking "otel_scope_name", "otel_scope_version" and "pod" but I don't mind them either if the value is stable

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                        self.assertIn(test_server.hostname, label_value)
+
+        # This tests whether each of the "byes sent" histogram type metric
+        # should have at least 1 data point whose mean should converge to be
+        # close to the number of bytes being sent by the RPCs.
+        with self.subTest("10_check_bytes_sent_vs_data_points"):
+            for metric in _CLIENT_SENT_METRICS:
+                self.assertTrue(
+                    self._at_least_one_point_within_range(
+                        all_results[metric][0].points, _REQUEST_PAYLOAD_SIZE
```
yashykt, line 476: note that with serialization, the actual value that you see would be slightly different, but I would expect to always see the same value.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                    ):
+                        # the "instance" label on the monitored resource refers
+                        # to the GKE pod and port
+                        self.assertIn(test_server.hostname, label_value)
+
+        # This tests whether each of the "byes sent" histogram type metric
+        # should have at least 1 data point whose mean should converge to be
+        # close to the number of bytes being sent by the RPCs.
+        with self.subTest("10_check_bytes_sent_vs_data_points"):
+            for metric in _CLIENT_SENT_METRICS:
```
yashykt, line 473: I'm not a fan of `_CLIENT_SENT_METRICS` being a list that contains `_METRIC_CLIENT_ATTEMPT_SENT,   _METRIC_SERVER_CALL_RCVD`. 

Just reading from the test, it does not make it obvious that the server rcvd metric is also being checked. I would just list both of them here.

Note this is also a matter of style, so I don't want this to block submission of the PR. I'm fine with it being resolved later.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                self.assertTrue(
+                    self._at_least_one_point_within_range(
+                        all_results[metric][0].points, _REQUEST_PAYLOAD_SIZE
+                    )
+                )
+
+            for metric in _SERVER_SENT_METRICS:
```
yashykt, line 480: same comment here as `_CLIENT_SENT_METRICS`

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                    "grpc_method",
+                    "otel_scope_name",
+                    "otel_scope_version",
+                    "pod",
+                ]:
+                    self.assertIn(label_key, metric_labels)
+
+        # Testing whether each metric has the right set of monitored resource
+        # label keys
+        with self.subTest("7_check_resource_label_keys"):
```
yashykt, line 302: I don't care too much about this either, but I'm fine with leaving it in.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +100,389 @@ def setUpClass(cls):
...
+                        label_key == "csm_workload_canonical_service"
+                        and self._is_server_metric(metric)
+                    ):
+                        # the server workload canonical service name
+                        self.assertEqual(
+                            _CSM_CANONICAL_SERVICE_NAME_SERVER, label_value
+                        )
+
+        # Testing the values of monitored resource labels
+        with self.subTest("9_check_resource_label_values"):
```
yashykt, line 421: i'm fine with the checks on the resource labels but again i don't care about them

tests/gamma/csm_observability_test.py
stanley-cheung: Removed checking against these keys.

tests/gamma/csm_observability_test.py
stanley-cheung: Yea I left a 5% margin in there, which should be good enough.

tests/gamma/csm_observability_test.py
stanley-cheung: Good point. I removed that array and just list the 2 metrics to be listed explicitly now. The array is kinda useless when it's used only in 1 place each.

tests/gamma/csm_observability_test.py
stanley-cheung: Doone

tests/gamma/csm_observability_test.py
stanley-cheung: I will leave it as is coz it helps verify the GMP part of things.

tests/gamma/csm_observability_test.py
stanley-cheung: Same as above.

yashykt (approved):

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+                            "label_value": label_value,
+                        },
+                    )
+                time_series.append(series)
+            results[metric] = time_series
+        return results
+
+    # A helper function to check whether at least one of the "points" whose
+    # mean should be within 5% of ref_bytes.
+    def _at_least_one_point_within_range(self, points, ref_bytes) -> bool:
```
sergiitk, line 137: Call it `assertAtLeastOnePointWithinRange` and assert directly in the method.
Then instead of 

```py
self.assertTrue(self._at_least_one_point_within_range(...))
```

you'll be able to do 

```py
self.assertAtLeastOnePointWithinRange(...)
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+                },
+                interval=interval,
+                view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+            )
+            time_series = []
+            logger.debug("Metric: %s", metric)
+            for series in response:
+                logger.debug(" Metric Labels:")
+                for label_key, label_value in series.metric.labels.items():
+                    logger.debug(
```
sergiitk, line 115: Instead of multiple `logger.debug()`, just make list, push each line into it, and do one 
```py
logger.debug('\n'.join(my_messages))
```

At this point we, can even do `logger.info`

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+                        # the remote workload canonical service name
+                        # for the server is the one we set in the client
+                        self.assertEqual(
+                            _CSM_CANONICAL_SERVICE_NAME_CLIENT, label_value
+                        )
+                    elif label_key == "csm_remote_workload_cluster_name":
+                        # the cluster name should be somewhere in the
+                        # kube context
+                        self.assertIn(
+                            label_value, xds_k8s_flags.KUBE_CONTEXT.value
```
sergiitk, line 337: Bad idea. There isn't no relation between the name of the cluster, and the name of the context. It just happens that gcloud generate context name in a certain way. But many choose to rename the context. In fact, my local run fail because of it:

```
E0126 16:05:28.531121 140704460725120 base_testcase.py:42] ----- PSM Test Case FAILED: CsmObservabilityTest.test_csm_observability -----
E0126 16:05:28.531229 140704460725120 base_testcase.py:95] (FAILURE) PSM Interop Test Failed: CsmObservabilityTest.test_csm_observability
^^^^^
[CsmObservabilityTest.test_csm_observability] PSM Failed Test Traceback BEGIN
Traceback (most recent call last):
  File "/Users/sergiitk/Development/psm-interop/tests/gamma/csm_observability_test.py", line 341, in test_csm_observability
    self.assertIn(
AssertionError: 'us-west1-a' not found in 'sergii-gamma'
[CsmObservabilityTest.test_csm_observability] PSM Failed Test Traceback END
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+                        )
+                    elif label_key == "csm_remote_workload_location":
+                        # the location should be somewhere in the kube context
+                        self.assertIn(
+                            label_value, xds_k8s_flags.KUBE_CONTEXT.value
```
sergiitk, line 342: same here

tests/gamma/csm_observability_test.py
```diff
@@ -29,6 +31,51 @@
...
+    "prometheus.googleapis.com/grpc_client_attempt_started_total/counter"
+)
+_METRIC_SERVER_CALL_RCVD = "prometheus.googleapis.com/grpc_server_call_rcvd_total_compressed_message_size_bytes/histogram"
+_METRIC_SERVER_CALL_SENT = "prometheus.googleapis.com/grpc_server_call_sent_total_compressed_message_size_bytes/histogram"
+_METRIC_SERVER_CALL_DURATION = (
+    "prometheus.googleapis.com/grpc_server_call_duration_seconds/histogram"
+)
+_METRIC_SERVER_CALL_STARTED = (
+    "prometheus.googleapis.com/grpc_server_call_started_total/counter"
+)
```
sergiitk, line 58:
```suggestion
_PROMETHEUS_HOST = "prometheus.googleapis.com"
_METRIC_CLIENT_ATTEMPT_SENT = (
    f"{_PROMETHEUS_HOST}/"
    "grpc_client_attempt_sent_total_compressed_message_size_bytes/histogram"
)
_METRIC_CLIENT_ATTEMPT_RCVD = (
    f"{_PROMETHEUS_HOST}/"
    "grpc_client_attempt_rcvd_total_compressed_message_size_bytes/histogram"
)
_METRIC_CLIENT_ATTEMPT_DURATION = (
    f"{_PROMETHEUS_HOST}/grpc_client_attempt_duration_seconds/histogram"
)
_METRIC_CLIENT_ATTEMPT_STARTED = (
    f"{_PROMETHEUS_HOST}/grpc_client_attempt_started_total/counter"
)
_METRIC_SERVER_CALL_RCVD = (
    f"{_PROMETHEUS_HOST}/"
    "grpc_server_call_rcvd_total_compressed_message_size_bytes/histogram"
)
_METRIC_SERVER_CALL_SENT = (
    f"{_PROMETHEUS_HOST}/"
    "grpc_server_call_sent_total_compressed_message_size_bytes/histogram"
)
_METRIC_SERVER_CALL_DURATION = (
    f"{_PROMETHEUS_HOST}/grpc_server_call_duration_seconds/histogram"
)
_METRIC_SERVER_CALL_STARTED = (
    f"{_PROMETHEUS_HOST}/grpc_server_call_started_total/counter"
)
```

tests/gamma/csm_observability_test.py
```diff
@@ -29,6 +31,51 @@
...
+_HISTOGRAM_CLIENT_METRICS = [
+    _METRIC_CLIENT_ATTEMPT_SENT,
+    _METRIC_CLIENT_ATTEMPT_RCVD,
+    _METRIC_CLIENT_ATTEMPT_DURATION,
+]
```
sergiitk, line 63: here and below: immutable lists (especially if constants) should be converted to tuples
```suggestion
_HISTOGRAM_CLIENT_METRICS = (
    _METRIC_CLIENT_ATTEMPT_SENT,
    _METRIC_CLIENT_ATTEMPT_RCVD,
    _METRIC_CLIENT_ATTEMPT_DURATION,
)
```

tests/gamma/csm_observability_test.py
```diff
@@ -29,6 +31,51 @@
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
 _Lang = skips.Lang
 
+# Testing consts
+_TEST_RUN_SECS = 90
```
sergiitk, line 35: let's drop `_` in all these consts. It's more distracting than helpful.

sergiitk (changes requested):
first round

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+            # 'grpc_status' label
+            filter_str = (
+                'metric.type = "%(metric)s" AND '
+                'metric.labels.grpc_method = "%(grpc_method)s"'
+            )
+            counter_results = self._query_metrics(
+                _COUNTER_METRICS, filter_str, interval
+            )
+
+            all_results = {**histogram_results, **counter_results}
```
sergiitk, line 218:
```suggestion
            all_results = {**histogram_results, **counter_results}
            self.assertNotEmpty(all_results, msg="No query metrics results")
```

If, for any reasons, returned results are empty, the rest of the tests that do assertions when iterating over `all_results` will falsely succeed.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +92,383 @@ def setUpClass(cls):
...
+
+        # Testing the values of metric labels
+        with self.subTest("8_check_metrics_label_values"):
+            for metric, time_series in all_results.items():
+                series = time_series[0]
+                self.assertEqual(metric, series.metric.type)
+                for label_key, label_value in series.metric.labels.items():
+                    if label_key == "pod" and self._is_client_metric(metric):
+                        # client pod name
+                        self.assertEqual(test_client.hostname, label_value)
```
sergiitk, line 305: I'm not too happy with how we make all these nested assertion loops with chained if-else. Not very readable. 
I think we should clearly break up server and client metrics in separate dictionaries, and check them separately. 
Then we'll be able to make use of `self.assertContainsSubset()` and `self.assertDictEqual()`.

tests/gamma/csm_observability_test.py
stanley-cheung: Done.

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: I added a sentinal value `VALUE_NOT_CHECKED` to indicate that we _do_ want to check whether the key exists, but we don't care about the value. This includes cluster_name, location, etc, for now.

tests/gamma/csm_observability_test.py
stanley-cheung: As above

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: I kept these as arrays for now because there are some `+` operations just slightly below. Is it OK?

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done. I also now explicitly assert that every metric should have _some_ results.

tests/gamma/csm_observability_test.py
stanley-cheung: Did a big refactor on this. No more loops checking each label_key and label value separately. Now for each metric, we are checking against a known bag of key: value pair. Some of the values can be `VALUE_NOT_CHECKED`, but for each check, all the values are in a single place now.

tests/gamma/csm_observability_test.py
```diff
@@ -25,13 +26,77 @@
...
+]
+COUNTER_SERVER_METRICS = [
+    METRIC_SERVER_CALL_STARTED,
+]
+HISTOGRAM_METRICS = HISTOGRAM_CLIENT_METRICS + HISTOGRAM_SERVER_METRICS
+COUNTER_METRICS = COUNTER_CLIENT_METRICS + COUNTER_SERVER_METRICS
+CLIENT_METRICS = HISTOGRAM_CLIENT_METRICS + COUNTER_CLIENT_METRICS
+SERVER_METRICS = HISTOGRAM_SERVER_METRICS + COUNTER_SERVER_METRICS
+ALL_METRICS = HISTOGRAM_METRICS + COUNTER_METRICS
+
```
sergiitk, line 94: I think we can make it even easier to understand with a dataclass to describe metric

```py
@dataclasses.dataclass(eq=False)
class Metric:
    name: str
    resource_labels: dict[str, str]
    metric_labels: dict[str, str]
    points: list[gm_metric.Point]

    @classmethod
    def from_response(
        cls,
        name: str,
        response: gm_metric.TimeSeries,
    ) -> "Metric":
        return cls(
            name=name,
            resource_labels=dict(response.metric.metric.labels),
            metric_labels=dict(response.metric.resource.labels),
            points=list(response.points),
        )
```

Let's discuss it IRL

sergiitk (changes requested):
Thanks for the last improvement! Much more readable already. Now, here's a few things to take care of

tests/gamma/csm_observability_test.py
```diff
@@ -25,13 +26,77 @@
...
+# Type aliases
 _XdsTestServer = xds_k8s_testcase.XdsTestServer
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
 _Lang = skips.Lang
 
+# Testing consts
+TEST_RUN_SECS = 90
+REQUEST_PAYLOAD_SIZE = 271828
+RESPONSE_PAYLOAD_SIZE = 314159
+VALUE_NOT_CHECKED = object()  # Sentinel value
```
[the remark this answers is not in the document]
sergiitk: A big upside of using `assertDictEqual` is the error will contain all fields:

```py
Traceback (most recent call last):
  File "/Users/sergiitk/Development/psm-interop/tests/gamma/csm_observability_test.py", line 380, in test_csm_observability
    self.assertDictEqual(
AssertionError: {'cluster': <ANY>, 'instance': <ANY>, 'job': 'fake me', 'location': <ANY>, 'namespace': 'sergiitk-client-20240130-0157-unzuy', 'project_id': 'sergiitk-grpc-gke'} != {'cluster': 'sergii-gamma', 'instance': 'psm-grpc-client-6cc57b7787-9sr4l:9464', 'job': 'psm-grpc-client-nlmj4oo8-gmp', 'location': 'us-west1-a', 'namespace': 'sergiitk-client-20240130-0157-unzuy', 'project_id': 'sergiitk-grpc-gke'}
repr() of differing entries:
'job': 'fake me' != 'psm-grpc-client-nlmj4oo8-gmp'
```

tests/gamma/csm_observability_test.py
stanley-cheung: Switched to tuple. The problem I had before probably had to do with the `b` in the example above only having 1 value, so that wasn't allowed:
```
>>> a = (1,2,3)
>>> b = (4)
>>> c = a + b
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate tuple (not "int") to tuple
```
Turns out i need to do this:
```
>>> b = (4,)
>>> c = a + b
>>> print(c)
(1, 2, 3, 4)
```

kubernetes-manifests/client.deployment.yaml
```diff
@@ -66,6 +66,22 @@ spec:
             value: "true"
           - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
             value: "true"
+          % if enable_csm_observability:
```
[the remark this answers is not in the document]
stanley-cheung: Done

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
+        logger.info("\n".join(metric_log_lines))
+        return results
+
+    # A helper function to check whether at least one of the "points" whose
+    # mean should be within 5% of ref_bytes.
+    def assertAtLeastOnePointWithinRange(self, points, ref_bytes):
+        for point in points:
+            if point.value.distribution_value.mean > (
+                ref_bytes * 0.95
+            ) and point.value.distribution_value.mean < (ref_bytes * 1.05):
```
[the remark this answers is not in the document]
stanley-cheung: Done

tests/gamma/csm_observability_test.py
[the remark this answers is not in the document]
stanley-cheung: Removed

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
+                return
+        self.fail(f"No data point with ~{ref_bytes} bytes found")
```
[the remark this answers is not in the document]
stanley-cheung: Done

tests/gamma/csm_observability_test.py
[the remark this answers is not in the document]
stanley-cheung: Thanks. `unittest.mock.ANY` is perfect. Now I am using `assertDictEqual` directly without any helper function. Works great!

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
+        )
+
     def test_csm_observability(self):
         # TODO(sergiitk): [GAMMA] Consider moving out custom gamma
         #   resource creation out of self.startTestServers()
         with self.subTest("1_run_test_server"):
-            test_server: _XdsTestServer = self.startTestServers(
-                enable_csm_observability=True
+            start_secs = int(time.time())
+            self.test_server = self.startTestServers(
```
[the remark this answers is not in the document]
stanley-cheung: It must have been an artifact after several refactors. I remember having to do this because I am making a separate function and need to pass in `test_client`. But after several refactors the need isn't there anymore. Removed all those.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
+                enable_csm_observability=True,
+                csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
+                csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
             )[0]
 
         with self.subTest("2_start_test_client"):
-            test_client: _XdsTestClient = self.startTestClient(
-                test_server,
+            self.test_client = self.startTestClient(
```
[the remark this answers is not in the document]
stanley-cheung: Done.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
+                    self.fail(f"Unknown metric: {metric}")
+
+                actual_metric_labels = all_results[metric][0].metric.labels
+                self.assertExpectedLabels(
+                    expected_metric_labels, actual_metric_labels
+                )
+
+        # Testing whether each metric has the right set of monitored resource
+        # label keys and values
+        with self.subTest("7_check_resource_labels"):
```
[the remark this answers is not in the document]
stanley-cheung: Done. Split them all up to make it less indented and less `if`

framework/test_app/runners/k8s/k8s_xds_client_runner.py
```diff
@@ -109,6 +109,8 @@ def run(  # pylint: disable=arguments-differ
         enable_csm_observability: bool = False,
         request_payload_size: int = 0,
         response_payload_size: int = 0,
+        csm_workload_name: str = "",
+        csm_canonical_service_name: str = "",
```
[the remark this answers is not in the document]
stanley-cheung: I am having a bit of trouble there. Will talk IRL

tests/gamma/csm_observability_test.py
sergiitk: Yea, it's a Python thing. Single-value tuples must be declared with a comma to distinguish from just in value in parens.

kubernetes-manifests/client.deployment.yaml
```diff
@@ -66,6 +66,24 @@ spec:
...
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: NAMESPACE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: OTEL_RESOURCE_ATTRIBUTES
+            value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
```
sergiitk, line 86: I think this block still should be within `% if enable_csm_observability:`. Let's talk IRL.

kubernetes-manifests/server.deployment.yaml
```diff
@@ -52,6 +52,24 @@ spec:
...
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: NAMESPACE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: OTEL_RESOURCE_ATTRIBUTES
+            value: k8s.pod.name=$(POD_NAME),k8s.namespace.name=$(NAMESPACE_NAME)
```
sergiitk, line 72: Same here.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +103,285 @@ def setUpClass(cls):
...
+            # server metrics are also serving on the Channelz requests.
+            #
+            filter_str_template = (
+                'metric.type = "%(metric)s" AND '
+                'metric.labels.grpc_status = "OK" AND '
+                f'metric.labels.grpc_method = "{GRPC_METHOD_NAME}"'
+            )
+            histogram_results = self.query_metrics(
+                HISTOGRAM_METRICS, filter_str_template, interval
+            )
```
sergiitk, line 207: here's how I think we should deal with the templating.

Move templates to parametrized methods.
```py
    @classmethod
    def _build_histogram_query(cls, metric_type: str) -> str:
       # Now you can also move that comment explaining query logic here
        return (
            f'metric.type = "{metric_type}" AND '
            'metric.labels.grpc_status = "OK" AND '
            f'metric.labels.grpc_method = "{GRPC_METHOD_NAME}"'
        )

    @classmethod
    def _build_counter_query(cls, metric_type: str) -> str:
        return (
            f'metric.type = "{metric_type}" AND '
            f'metric.labels.grpc_method = "{GRPC_METHOD_NAME}"'
        )
```

Make method reference an argument of `query_metrics`:

```py
BuildQueryFn = Callable[[str], str]
# ...
    def query_metrics(
        self,
        metric_names: Iterable[str],
        build_query_fn: BuildQueryFn,
        interval: monitoring_v3.TimeInterval,
    ):
# ...
         response = self.metric_client.list_time_series(
                name=f"projects/{self.project}",
                filter=build_query_fn(metric),
                interval=interval,
                view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
            )
```

Update the calls:

```py
            histogram_results = self.query_metrics(
                HISTOGRAM_METRICS, self._build_histogram_query, interval
            )
            counter_results = self.query_metrics(
                COUNTER_METRICS, self._build_histogram_query, interval
            )
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +103,285 @@ def setUpClass(cls):
...
+                time_series = all_results[metric]
+                # There should be exactly 1 time series per metric with these
+                # label value combinations that we are expecting.
+                self.assertEqual(1, len(time_series))
+
+        # Testing whether each metric has the correct set of metric keys and
+        # values
+        with self.subTest("6_check_metrics_labels"):
+            for metric in HISTOGRAM_CLIENT_METRICS:
+                expected_metric_labels = {
```
sergiitk, line 236: I don't think you need to define `expected_metric_labels` inside the loop now

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +103,285 @@ def setUpClass(cls):
...
+                actual_metric_labels = all_results[metric][0].metric.labels
+                self.assertDictEqual(
+                    expected_metric_labels, dict(actual_metric_labels)
+                )
+
+        # Testing whether each metric has the correct set of metric keys and
+        # values
+        with self.subTest("7_check_metrics_labels"):
+            for metric in HISTOGRAM_SERVER_METRICS:
+                expected_metric_labels = {
```
sergiitk, line 265: Same here

tests/gamma/csm_observability_test.py
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import time
+import unittest
```
sergiitk, line 16: Since mock.py is a file within unittest (which makes it a module),  `import unittest.mock` should be used
image

Honestly, I'm surprised it even worked. Probably a side-effect of something. Normally, it'll fail:

```pycon
In [1]: import unittest

In [2]: type(unittest)
Out[2]: module

In [3]: type(unittest.mock)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 type(unittest.mock)

File /opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/__init__.py:95, in __getattr__(name)
     93     from .async_case import IsolatedAsyncioTestCase
     94     return IsolatedAsyncioTestCase
---> 95 raise AttributeError(f"module {__name__!r} has no attribute {name!r}")

AttributeError: module 'unittest' has no attribute 'mock'

In [4]: import unittest.mock

In [5]: type(unittest.mock)
Out[5]: module
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +103,285 @@ def setUpClass(cls):
...
+                actual_metric_labels = all_results[metric][0].metric.labels
+                self.assertDictEqual(
+                    expected_metric_labels, dict(actual_metric_labels)
+                )
+
+        # Testing whether each metric has the correct set of metric keys and
+        # values
+        with self.subTest("8_check_metrics_labels"):
+            for metric in COUNTER_CLIENT_METRICS:
+                expected_metric_labels = {
```
sergiitk, line 291: same here and elsewhere below

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +103,285 @@ def setUpClass(cls):
...
+                actual_resource_labels = time_series.resource.labels
+                self.assertDictEqual(
+                    expected_resource_labels, dict(actual_resource_labels)
+                )
+
+        # This tests whether each of the "byes sent" histogram type metric
+        # should have at least 1 data point whose mean should converge to be
+        # close to the number of bytes being sent by the RPCs.
+        with self.subTest("12_check_bytes_sent_vs_data_points"):
+            for metric in [
```
sergiitk, line 370: nit: make it a tuple too

sergiitk (changes requested):
Almost there!

kubernetes-manifests/client.deployment.yaml
stanley-cheung: Done. Makes sense to me.

kubernetes-manifests/server.deployment.yaml
stanley-cheung: Done.

tests/gamma/csm_observability_test.py
stanley-cheung: Done. This is great, much more readable this way. Thanks for code snippets.

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed.

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Working on this.

sergiitk: Just FYI: I wasn't 100% sure this structure works for you, so I wanted to discuss this IRL first.

framework/test_app/runners/k8s/k8s_xds_client_runner.py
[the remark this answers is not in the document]
sergiitk: Made an example for the server runner:

```diff
diff --git a/framework/test_app/runners/k8s/gamma_server_runner.py b/framework/test_app/runners/k8s/gamma_server_runner.py
index cf0ffcd8b5..6394e8a888 100644
--- a/framework/test_app/runners/k8s/gamma_server_runner.py
+++ b/framework/test_app/runners/k8s/gamma_server_runner.py
@@ -46,2 +46,4 @@ class GammaServerRunner(KubernetesServerRunner):
     frontend_service_name: str
+    csm_workload_name: str
+    csm_canonical_service_name: str
 
@@ -77,2 +79,4 @@ class GammaServerRunner(KubernetesServerRunner):
         pre_stop_hook: bool = False,
+        csm_workload_name: str = "",
+        csm_canonical_service_name: str = "",
     ):
@@ -109,2 +113,4 @@ class GammaServerRunner(KubernetesServerRunner):
         self.pre_stop_hook = pre_stop_hook
+        self.csm_workload_name = csm_workload_name
+        self.csm_canonical_service_name = csm_canonical_service_name
 
@@ -210,4 +216,4 @@ class GammaServerRunner(KubernetesServerRunner):
             generate_mesh_id=generate_mesh_id,
-            csm_workload_name=csm_workload_name,
-            csm_canonical_service_name=csm_canonical_service_name,
+            csm_workload_name=self.csm_workload_name,
+            csm_canonical_service_name=self.csm_canonical_service_name,
         )
diff --git a/framework/xds_gamma_testcase.py b/framework/xds_gamma_testcase.py
index b535b9ffba..fc8d927331 100644
--- a/framework/xds_gamma_testcase.py
+++ b/framework/xds_gamma_testcase.py
@@ -96,3 +96,3 @@ class GammaXdsKubernetesTestCase(xds_k8s_testcase.RegularXdsKubernetesTestCase):
 
-    def initKubernetesServerRunner(self) -> GammaServerRunner:
+    def initKubernetesServerRunner(self, **kwargs) -> GammaServerRunner:
         return GammaServerRunner(
@@ -114,2 +114,3 @@ class GammaXdsKubernetesTestCase(xds_k8s_testcase.RegularXdsKubernetesTestCase):
             pre_stop_hook=self.pre_stop_hook,
+            **kwargs,
         )
diff --git a/tests/gamma/csm_observability_test.py b/tests/gamma/csm_observability_test.py
index 28ec344ad9..1bc793447e 100644
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -25,2 +25,3 @@ from framework import xds_k8s_testcase
 from framework.helpers import skips
+from framework.test_app.runners.k8s import gamma_server_runner
 
@@ -90,2 +91,4 @@ ALL_METRICS = HISTOGRAM_METRICS + COUNTER_METRICS
 
+GammaServerRunner = gamma_server_runner.GammaServerRunner
+
 
@@ -191,4 +194,2 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
                 enable_csm_observability=True,
-                csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
-                csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
             )[0]
@@ -385,2 +386,8 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
 
+    def initKubernetesServerRunner(self) -> GammaServerRunner:
+        return super().initKubernetesServerRunner(
+            csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
+        )
+
```

tests/gamma/csm_observability_test.py
stanley-cheung: Added the `MetricTimeSeries` `dataclass` here. Please review. Thanks. It has a `pretty_print()` method. In the rest of the test, when we access `all_results[metric]` we know we are getting a `MetricTimeSeries` back.

framework/test_app/runners/k8s/k8s_xds_client_runner.py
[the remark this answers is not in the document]
stanley-cheung: Ah right on. This piece

```
-    def initKubernetesServerRunner(self) -> GammaServerRunner:
+    def initKubernetesServerRunner(self, **kwargs) -> GammaServerRunner:
```

was the key piece that I wasn't sure on. Now I get what you mean.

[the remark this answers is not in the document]
stanley-cheung: Done! Now these 2 parameters are not passed as `run()` parameters anymore. Instead they are passed to the constructor of the client runner and the server runner.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+                self.assertEqual(
+                    "prometheus_target", metric_time_series.resource_type
+                )
+
+                actual_resource_labels = metric_time_series.resource_labels
+                self.assertDictEqual(
+                    expected_resource_labels, actual_resource_labels
+                )
+
+        # This tests whether each of the "byes sent" histogram type metric
```
sergiitk, line 421: bytes sent?

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+        # should have at least 1 data point whose mean should converge to be
+        # close to the number of bytes being sent by the RPCs.
+        with self.subTest("12_check_bytes_sent_vs_data_points"):
+            for metric in (
+                METRIC_CLIENT_ATTEMPT_SENT,
+                METRIC_SERVER_CALL_RCVD,
+            ):
```
sergiitk, line 428: Nit: this fits into the line length now
```suggestion
            for metric in (METRIC_CLIENT_ATTEMPT_SENT, METRIC_SERVER_CALL_RCVD):
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+                self.assertAtLeastOnePointWithinRange(
+                    all_results[metric].points, REQUEST_PAYLOAD_SIZE
+                )
+
+            for metric in (
+                METRIC_CLIENT_ATTEMPT_RCVD,
+                METRIC_SERVER_CALL_SENT,
+            ):
```
sergiitk, line 436: same
```suggestion
            for metric in (METRIC_CLIENT_ATTEMPT_RCVD, METRIC_SERVER_CALL_SENT):
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+
+    # These parameters are more pertaining to the test itself, not to
+    # each run().
+    def initKubernetesServerRunner(self, **kwargs) -> GammaServerRunner:
+        return super().initKubernetesServerRunner(
+            csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
+        )
+
+    BuildQueryFn = Callable[[str], str]
```
sergiitk, line 166: This should be defined next to other type aliases (lines 95-96)

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+                ref_bytes * (1 - tolerance)
+                < point.value.distribution_value.mean
+                < ref_bytes * (1 + tolerance)
+            ):
+                return
+        self.fail(
+            f"No data point with {ref_bytes}±{tolerance*100}% bytes found"
+        )
+
     def test_csm_observability(self):
```
sergiitk, line 245: Normally we want to place the test above other helper methods, but after method overrides. Let's move it up and place it after `initKubernetesServerRunner`.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
-                        name="whatever",
-                    )
+        with self.subTest("4_query_cloud_monitoring_metrics"):
+            end_secs = int(time.time())
+            interval = monitoring_v3.TimeInterval(
+                {
+                    "end_time": {"seconds": end_secs, "nanos": 0},
+                    "start_time": {"seconds": start_secs, "nanos": 0},
+                }
+            )
```
sergiitk, line 274: nit
```suggestion
            interval = monitoring_v3.TimeInterval(
                start_time={"seconds": start_secs},
                end_time={"seconds": end_secs},
            )
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+        with self.subTest("5_check_metrics_time_series"):
+            for metric in ALL_METRICS:
+                # Every metric needs to exist in the query results
+                self.assertIn(metric, all_results)
+
+        # Testing whether each metric has the correct set of metric keys and
+        # values
+        with self.subTest("6_check_metrics_labels_histogram_client"):
+            expected_metric_labels = {
+                "csm_mesh_id": unittest.mock.ANY,
```
sergiitk, line 293: Nit: I think we can make it significantly less cluttered and easier to read by defining ANY in the type alias block  (lines 95-96) as

```
ANY = unittest.mock.ANY
```

And replacing any in the expected dicts.

tests/gamma/csm_observability_test.py
```diff
@@ -11,24 +11,126 @@
...
+
+GammaServerRunner = gamma_server_runner.GammaServerRunner
+KubernetesClientRunner = k8s_xds_client_runner.KubernetesClientRunner
+
+
+# This class should represent one TimeSeries object from
+# monitoring_v3.ListTimeSeriesResponse.
+@dataclasses.dataclass(eq=False)
+class MetricTimeSeries:
+    # the metric name
```
sergiitk, line 103: Looks like a great candidate to be placed in the class doctring

```suggestion
@dataclasses.dataclass(eq=False)
class MetricTimeSeries:
    """
    This class represents one TimeSeries object
    from monitoring_v3.ListTimeSeriesResponse.
    """

    # the metric name
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+    def build_counter_query(cls, metric_type: str) -> str:
+        # For these num rpcs started counter metrics, they do not have the
+        # 'grpc_status' label
+        return (
+            f'metric.type = "{metric_type}" AND '
+            f'metric.labels.grpc_method = "{GRPC_METHOD_NAME}"'
+        )
+
+    # A helper function to make the cloud monitoring API call to query metrics
+    # created by this test run.
```
sergiitk, line 198: Make it a docstring.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+                )
+            metric_time_series = MetricTimeSeries.from_response(
+                metric, time_series[0]
+            )
+            logger.info(metric_time_series.pretty_print())
+            results[metric] = metric_time_series
+        return results
+
+    # A helper function to check whether at least one of the "points" whose
+    # mean should be within 5% of ref_bytes.
```
sergiitk, line 227: Make it a docstring.

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+    # created by this test run.
+    def query_metrics(
+        self,
+        metric_names: Iterable[str],
+        build_query_fn: BuildQueryFn,
+        interval: monitoring_v3.TimeInterval,
+    ) -> dict[str, MetricTimeSeries]:
+        results = {}
+        for metric in metric_names:
+            response = self.metric_client.list_time_series(
```
sergiitk, line 207: Since this makes an external request, we want to make users know what's happening
```suggestion
            logger.info("Requesting list_time_series for metric %s", metric)
            response = self.metric_client.list_time_series(
```

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+            time_series = list(response)
+            if len(time_series) > 1:
+                self.fail(
+                    f"Query for {metric} should return exactly 1 time series. "
+                    f"Found {len(time_series)}."
+                )
+            metric_time_series = MetricTimeSeries.from_response(
+                metric, time_series[0]
+            )
+            logger.info(metric_time_series.pretty_print())
```
sergiitk, line 222: Just to make it a little nicer:
```suggestion
            logger.info(
                "Metric %s:\n%s", metric, metric_time_series.pretty_print()
            )
```

tests/gamma/csm_observability_test.py
```diff
@@ -11,24 +11,126 @@
...
+    def from_response(
+        cls,
+        name: str,
+        response: monitoring_v3.types.TimeSeries,
+    ) -> "MetricTimeSeries":
+        return cls(
+            name=name,
+            metric_labels=dict(response.metric.labels),
+            resource_type=response.resource.type,
+            resource_labels=dict(response.resource.labels),
```
sergiitk, line 124: Since proto maps are always out of order, it may be simpler for us to presort the keys
```suggestion
            metric_labels=dict(sorted(response.metric.labels.items())),
            resource_type=response.resource.type,
            resource_labels=dict(sorted(response.resource.labels.items())),
```

tests/gamma/csm_observability_test.py
```diff
@@ -11,24 +11,126 @@
...
+            points=list(response.points),
+        )
+
+    def pretty_print(self) -> str:
+        metric = dataclasses.asdict(self)
+        # too much noise to print all data points from a time series
+        metric.pop("points")
+        return yaml.dump(metric, sort_keys=True)
```
sergiitk, line 132: If we go with the presort I suggested above, we may benefit from printing while preserving the top-level fields order (name, resource_type, resource_labels, etc). If so, would be nice to place `resource_type: str` right next to the name

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +147,296 @@ def setUpClass(cls):
...
+        results = {}
+        for metric in metric_names:
+            response = self.metric_client.list_time_series(
+                name=f"projects/{self.project}",
+                filter=build_query_fn(metric),
+                interval=interval,
+                view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
+            )
+            time_series = list(response)
+            if len(time_series) > 1:
```
sergiitk, line 214:
```suggestion
            if len(time_series) != 1:
```
I guess to cover for the case when the len is 0

sergiitk, line 214:
```suggestion
            if len(time_series) != 1:
```
I guess to cover for the case when the len is 0

sergiitk (commented):
I really like the changes you made! It looks great! Thank you so much for making this test so much more readable and maintainable!
A few cosmetic & style improvements, and we're good to go!

tests/gamma/csm_observability_test.py
sergiitk: I think the system posted this comment two times.

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed.

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Moved.

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed

tests/gamma/csm_observability_test.py
```diff
@@ -45,39 +110,264 @@ def setUpClass(cls):
...
                 enable_csm_observability=True,
-                request_payload_size=271828,
-                response_payload_size=314159,
+                request_payload_size=REQUEST_PAYLOAD_SIZE,
+                response_payload_size=RESPONSE_PAYLOAD_SIZE,
+                csm_workload_name=CSM_WORKLOAD_NAME_CLIENT,
+                csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_CLIENT,
             )
+            logger.info("Letting test client run for %d seconds", TEST_RUN_SECS)
+            time.sleep(TEST_RUN_SECS)
```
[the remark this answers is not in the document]
sergiitk: Yes, I think it should. If Initial RPCs fail, we won't be waiting extra 90 seconds.

sergiitk (commented):
Running tests:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/f628e63b-4382-4264-ab43-faaa63b3d028)
- [ ] [grpc/java/v1.61.x/branch/psm-security](https://source.cloud.google.com/results/invocations/34aaa0ae-a2a8-4363-94b3-c103e8cfaf25)
- [ ] [grpc/java/v1.61.x/branch/xds_k8s_lb](https://source.cloud.google.com/results/invocations/4896e665-369d-493d-94e5-6c73b3bc2118)
- [x] [grpc/java/v1.61.x/branch/xds_url_map](https://source.cloud.google.com/results/invocations/2ffd18cc-6b47-419d-a1cb-454fbe42633c)

PR #24, opened 2024-01-23
Add --generate-mesh-id-experimental flag to server bootstrap

stanley-cheung:
This flag has been added to the client deployment but not the server deployment. Adding it to server now.

We need this fix in order for the CSM Observability Test to properly record the server side metrics.

### DIFF
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -50,7 +50,8 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
         #   resource creation out of self.startTestServers()
         with self.subTest("1_run_test_server"):
             test_server: _XdsTestServer = self.startTestServers(
-                enable_csm_observability=True
+                enable_csm_observability=True,
+                generate_mesh_id=True,
             )[0]
 
         with self.subTest("2_start_test_client"):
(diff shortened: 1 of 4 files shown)


tests/gamma/csm_observability_test.py
```diff
@@ -50,7 +50,8 @@ def test_csm_observability(self):
         #   resource creation out of self.startTestServers()
         with self.subTest("1_run_test_server"):
             test_server: _XdsTestServer = self.startTestServers(
-                enable_csm_observability=True
+                enable_csm_observability=True,
```
gnossen, line 53: Should we add this to an override of [`RegularXdsKubernetesTestCase.startTestServers`](https://github.com/grpc/psm-interop/blob/292cea23d6263862416008b0a76b4c0c16cbaf69/framework/xds_k8s_testcase.py#L834) within [`GammaXdsKubernetesTestCase`](https://github.com/grpc/psm-interop/blob/292cea23d6263862416008b0a76b4c0c16cbaf69/framework/xds_gamma_testcase.py#L35), similar to how `startTestClient` currently has a GAMMA-specific override. This way, all future CSM tests will get this and we won't have to go through this debugging process again in the future.

sergiitk: Good point - but do we actually need to set `generate_mesh_id` in all gamma tests? If so, then we should make `GammaXdsKubernetesTestCase.startTestServers()` to set it automatically. If we don't (f.e. in affinity / baseline) tests, I think we should keep it as is.

gnossen: Yes. We do need `generate_mesh_id` in all GAMMA tests.

stanley-cheung: Ok, will try to refactor. Thanks for the confirmation.

sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/8665b61c-4627-41cf-b0bb-71fd08689694)

sergiitk (approved):
Refactoring done, all good after the test passes

PR #26, opened 2024-01-24
Fix bug with Gamma Server Runner not having GCP Logs link

stanley-cheung:
This fixes a bug with all the `GammaServerRunner` subclass not calling the super class `KubernetesBaseRunner`'s `run()` method properly, which causes the variable `self.time_start_requested` not being set and hence there is never `RunHistory` logged for the gamma server runner.

ref b/298501683

### DIFF
--- a/framework/test_app/runners/k8s/gamma_server_runner.py
+++ b/framework/test_app/runners/k8s/gamma_server_runner.py
@@ -20,6 +20,7 @@ from typing import List, Optional
 
 from framework.infrastructure import gcp
 from framework.infrastructure import k8s
+from framework.test_app.runners.k8s import k8s_base_runner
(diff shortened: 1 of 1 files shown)


framework/test_app/runners/k8s/gamma_server_runner.py
sergiitk, line 1: - [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/173185a7-0eee-4b8a-8da3-f23474c71e37)

sergiitk (approved):

framework/test_app/runners/k8s/gamma_server_runner.py
stanley-cheung: This job has also passed!

PR #27, opened 2024-02-01
CSM Observability: update cpp version check to v1.62.x

sergiitk:
Until (and if) all changes needed for CSM Observability are backported to `v1.61.x`, we should only allow this test for the  upcoming release `v1.62.x`, and the `master` branch.


sergiitk:
cc @stanley-cheung

gnossen (approved):

PR #29, opened 2024-02-02
Reorder CSM o11y test function order 

stanley-cheung:
… to place main test function near the top. Missed in #23.

### DIFF
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -152,109 +152,6 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
         super().setUpClass()
         cls.metric_client = cls.gcp_api_manager.monitoring_metric_service("v3")
 
-    # These parameters are more pertaining to the test itself, not to
-    # each run().
-    def initKubernetesClientRunner(self, **kwargs) -> KubernetesClientRunner:
-        return super().initKubernetesClientRunner(
-            csm_workload_name=CSM_WORKLOAD_NAME_CLIENT,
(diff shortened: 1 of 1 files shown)


tests/gamma/csm_observability_test.py
sergiitk, line 1: Running the test. Feel free to resolve once it passes to auto-merge the PR.

- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/9304a179-814b-4dc5-9391-df021aa43601)

sergiitk (approved):

tests/gamma/csm_observability_test.py
```diff
@@ -446,6 +343,109 @@ def test_csm_observability(self):
...
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_CLIENT,
+        )
+
+    # These parameters are more pertaining to the test itself, not to
+    # each run().
+    def initKubernetesServerRunner(self, **kwargs) -> GammaServerRunner:
+        return super().initKubernetesServerRunner(
+            csm_workload_name=CSM_WORKLOAD_NAME_SERVER,
+            csm_canonical_service_name=CSM_CANONICAL_SERVICE_NAME_SERVER,
+        )
```
sergiitk, line 360: We should keep these two at the top, after `setUpClass`, but before the test. These are method overrides, and are not explicitly called from the test. That's why I want to keep them visible.

sergiitk (changes requested):

tests/gamma/csm_observability_test.py
stanley-cheung: Fixed.

sergiitk (approved):

PR #32, opened 2024-02-07
[CSM O11y test] Filter out time series with no valid data point

stanley-cheung:
- From a prod test failure, we observed that there may be time series data in a metric that contain no valid data points. It's unclear why they come to be at the moment
- This PR attempts to not count time series when there is no valid data point in the time series

Can't really do the filtering at the query level because the monitoring API does not support filtering by data points: https://cloud.google.com/monitoring/api/v3/filters

### DIFF
--- a/tests/gamma/csm_observability_test.py
+++ b/tests/gamma/csm_observability_test.py
@@ -429,10 +429,7 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
                 view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
                 retry=retry_settings,
             )
-            time_series = []
-            for series in response:
-                if self.is_legit_time_series(series):
-                    time_series.append(series)
+            time_series = list(filter(self.is_legit_time_series, response))
 
             self.assertLen(
                 time_series,
@@ -450,20 +447,19 @@ class CsmObservabilityTest(xds_gamma_testcase.GammaXdsKubernetesTestCase):
             results[metric] = metric_time_series
         return results
 
+    @classmethod
     def is_legit_time_series(
-        self, series: monitoring_v3.types.TimeSeries
+        cls, series: monitoring_v3.types.TimeSeries
     ) -> bool:
         for point in series.points:
             # Test flake: we found some time series with no actual data point
             # in prod test runs.
             # Here we will only include time series with actual data in it.
-            if (
-                point.value.distribution_value.count > 0
-                or point.value.double_value > 0
-            ):
+            if point.value.distribution_value.count or point.value.double_value:
                 return True
-        logger.error("Warning: found time_series with no valid data point")
-        logger.error(series)
+        logger.warning(
+            "Warning: found time_series with no valid data point\n%s", series
+        )
         return False
 
     def assertAtLeastOnePointWithinRange(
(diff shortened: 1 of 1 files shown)


stanley-cheung:
Note that this did not really address the root cause as in _why are there an extra time series with no valid data point_.

tests/gamma/csm_observability_test.py
```diff
@@ -447,6 +450,22 @@ def query_metrics(
             results[metric] = metric_time_series
         return results
 
+    def is_legit_time_series(
+        self, series: monitoring_v3.types.TimeSeries
+    ) -> bool:
```
sergiitk, line 455: This method doesn't really need object state, let's make it a `@classmethod`

tests/gamma/csm_observability_test.py
```diff
@@ -447,6 +450,22 @@ def query_metrics(
...
+        for point in series.points:
+            # Test flake: we found some time series with no actual data point
+            # in prod test runs.
+            # Here we will only include time series with actual data in it.
+            if (
+                point.value.distribution_value.count > 0
+                or point.value.double_value > 0
+            ):
+                return True
+        logger.error("Warning: found time_series with no valid data point")
```
sergiitk, line 465: `logger.warning`

tests/gamma/csm_observability_test.py
```diff
@@ -447,6 +450,22 @@ def query_metrics(
...
+        logger.error(series)
```
sergiitk, line 466: merge with previous logger statement

tests/gamma/csm_observability_test.py
```diff
@@ -447,6 +450,22 @@ def query_metrics(
...
+        self, series: monitoring_v3.types.TimeSeries
+    ) -> bool:
+        for point in series.points:
+            # Test flake: we found some time series with no actual data point
+            # in prod test runs.
+            # Here we will only include time series with actual data in it.
+            if (
+                point.value.distribution_value.count > 0
+                or point.value.double_value > 0
+            ):
```
sergiitk, line 463: Pythonistic fix :)
```suggestion
            if point.value.distribution_value.count or point.value.double_value:
                return True
```

tests/gamma/csm_observability_test.py
```diff
@@ -429,7 +429,10 @@ def query_metrics(
                 view=monitoring_v3.ListTimeSeriesRequest.TimeSeriesView.FULL,
                 retry=retry_settings,
             )
-            time_series = list(response)
+            time_series = []
+            for series in response:
+                if self.is_legit_time_series(series):
+                    time_series.append(series)
```
sergiitk, line 435: Optional: can be simplified to 
```suggestion
            time_series = list(filter(self.is_legit_time_series, response))
```

sergiitk (changes requested):

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
stanley-cheung: Done

tests/gamma/csm_observability_test.py
sergiitk, line 1: Resolve the conversation once the test pass to merge the PR:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/7c2ea706-8a64-4c44-af92-04c5d190d400)

sergiitk (approved):
All good - will enable PR auto-merge, but block on the unresolved conversation below.

tests/gamma/csm_observability_test.py
stanley-cheung: Job passed

PR #40, opened 2024-02-16
[CSM O11y] Avoid collisions between test runs when querying metrics

yashykt:
We noticed a bug in CSM O11y test harness where we did not scope our metric queries to a particular test run. The fix is to add a scope on the client/server namespace which should be unique to a specific test run.


sergiitk:
- [ ] [grpc/core/v1.62.x/branch/linux/psm-csm](https://source.cloud.google.com/results/invocations/78d74ab5-139f-4124-9dd4-1aef09fb5a53)

sergiitk (approved):
Looks great! Ready to be merged once the adhoc run passes

stanley-cheung:
Thank you for the fix.

PR #41, opened 2024-02-20
Fix the issue with Crtl+C infinite loop during the cleanup

sergiitk:
Fixes the bug with pressing Ctrl+C second time during test teardown causing the teardown to start over, in an infinite loop.
Now second Ctrl+C will abort the cleanup and exit immediately. Also, it gives a chance to press Ctrl+C second time before any cleanup started by pausing before the cleanup for 2 seconds.


sergiitk:
FYI @stanley-cheung

XuanWang-Amos (approved):

PR #43, opened 2024-02-22
[Gamma] Bump the timeout for wait_service_neg_status_annotation to 4 mins

arvindbr8:
Gamma/CSM tests: 99 percentile of tests are passing, while we see a few failures in neg status annotation propagation on the service. The dashboards do not show anything strange. Bumping the timeout to 4 mins.

Ref b/298501683, b/298501683


sergiitk (approved):
Thanks for the change!

PR #45, opened 2024-02-27
Style: add typing-extensions per updated Google Python Style Guide 

sergiitk:
[Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) has evolved, and looks all modern and futuristic now. I'll be getting tolling in this repo closer to match closer to it.

- Add [`typing-extensions`](https://typing-extensions.readthedocs.io/en/latest/) package.
- Configure isort single_line_exclusions according to https://google.github.io/styleguide/pyguide.html#2241-exemptions

With `typing-extensions`, we can use such cool things as [`TypeAlias`](https://docs.python.org/3/library/typing.html#typing.TypeAlias), [`Self`](https://docs.python.org/3/library/typing.html#typing.Self), and [`@override`](https://docs.python.org/3/library/typing.html#typing.override).

Note: we don't migrate the code to the new types yet, but we should be using these features when writing new code.

### DIFF
--- a/.github/workflows/psm-interop.yaml
+++ b/.github/workflows/psm-interop.yaml
@@ -72,7 +72,7 @@ jobs:
         - uses: actions/checkout@v4
         - uses: isort/isort-action@v1
           with:
-            isort-version: "5.9.2"
+            isort-version: "5.13.2"
 
   pylint-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


sergiitk:
cc @gnossen

XuanWang-Amos (approved):

PR #49, opened 2024-03-04
Improve port forwarding logging

sergiitk:
Makes it more clear to what pod port forwarding started - both when starting and shutting down port forwarding.

Starting:
```
I0304 15:00:51.754341 140704470529984 k8s.py:1023] LOCAL DEV MODE: Enabling port forwarding to pod/psm-grpc-server-79cdbdddc7-w668w 10.118.1.122:8080
I0304 15:00:52.664326 140704470529984 k8s_port_forwarder.py:118] Forwarding from 127.0.0.1:61903 -> 8080
```

Shutting down:
```
I0304 15:04:21.526046 140704470529984 k8s_port_forwarder.py:126] Shutting down PortForwarder(pid='58289', namespace='sergiitk-server-20240304-2300-1db6t', destination='pod/psm-grpc-server-79cdbdddc7-w668w', '127.0.0.1:61903 -> 8080')
I0304 15:04:21.529421 140704470529984 k8s_base_runner.py:201] Port forwarders in namespace sergiitk-server-20240304-2300-1db6t stopped.
```

### DIFF
--- a/framework/test_app/runners/k8s/k8s_base_runner.py
+++ b/framework/test_app/runners/k8s/k8s_base_runner.py
@@ -193,8 +193,13 @@ class KubernetesBaseRunner(base_runner.BaseRunner, metaclass=ABCMeta):
         self.maybe_stop_logging()
 
         # Stop port forwarders if any.
-        for pod_port_forwarder in self.pod_port_forwarders:
-            pod_port_forwarder.close()
+        if self.pod_port_forwarders:
+            for pod_port_forwarder in self.pod_port_forwarders:
+                pod_port_forwarder.close()
+
+            ns = self.k8s_namespace.name if self.k8s_namespace else "Unknown"
+            logger.info("Port forwarders in namespace %s stopped.", ns)
+
         self.pod_port_forwarders = []
 
         for pod_log_collector in self.pod_log_collectors:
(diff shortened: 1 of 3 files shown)


framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -193,8 +193,13 @@ def stop_pod_dependencies(self, *, log_drain_sec: int = 0):
         self.maybe_stop_logging()
 
         # Stop port forwarders if any.
-        for pod_port_forwarder in self.pod_port_forwarders:
-            pod_port_forwarder.close()
+        if self.pod_port_forwarders:
+            for pod_port_forwarder in self.pod_port_forwarders:
```
XuanWang-Amos, line 197: [Minor] I don't think we need an `if` here, `for..in..` should be able to handle empty list.

XuanWang-Amos (approved):

framework/test_app/runners/k8s/k8s_base_runner.py
sergiitk: Yea - I agree. This is purely for the log statement below. I used to have it just for the log statement, but then logically it made more sense to group both for loop and logging under the same if.

sergiitk: Fixed

PR #48, opened 2024-03-04
Fix incorrect dependency on xds_url_map_testcase for client config logic

sergiitk:
Some tests incorrectly import `framework.xds_url_map_testcase` to get `UNARY_CALL` , `EMPTY_CALL` constants. This should have never been the case: csds logic and helper should be `framework.rpc.grpc_testing`.

This PR:
1. Moves `xds_url_map_testcase.RpcType*Call` to `grpc_testing.RPC_TYPE_*_CALL`.
2. Adds `configure_unary`, `configure_empty`, `configure_rpc_type` helpers to XdsUpdateClientConfigureServiceClient.
3. Adds `RPC_TYPES_BOTH_CALLS` constant import to simplify method calls where both call types are expected.
4. Updates all tests to import constants directly from `grpc_testing`.

Similar to https://github.com/grpc/psm-interop/pull/47.

PR #47, opened 2024-03-04
Fix incorrect dependency on xds_url_map_testcase for CSDS logic

sergiitk:
Some tests incorrectly on `framework.xds_url_map_testcase` to get `DumpedXdsConfig`. This should have never been the case: csds logic and helper should be `framework.rpc.grpc_csds`.

This is especially wrong in `framework.xds_k8s_testcase` depending on `framework.xds_url_map_testcase`, while `framework.xds_url_map_testcase` depends on `framework.xds_k8s_testcase`. I'm surprised this didn't cause more problems.

This PR:
1. Moves `DumpedXdsConfig` to `framework.rpc.grpc_csds`
2. Refactors `DumpedXdsConfig` for readability
3. Addresses TODO wrt catching broad exception while parsing
4. Adds CsdsClient.fetch_client_status_parsed that returns `DumpedXdsConfig` so class users don't need to depend on json
5. Replaces all imports of `xds_url_map_testcase` that were needed just for `DumpedXdsConfig` with imports of `grpc_csds`

PR #51, opened 2024-03-05
Fix incorrect dependency on xds_url_map_testcase for RpcDistributionStats

sergiitk:
Same as #47, #48, but for `RpcDistributionStats`.

This concludes the fix for xds_url_map_testcase dependency; after this change only url_map tests depend on it.

### DIFF
--- a/framework/rpc/grpc_testing.py
+++ b/framework/rpc/grpc_testing.py
@@ -78,6 +78,7 @@ class RpcDistributionStats:
     empty_call_alternative_service_rpc_count: int
 
     def __init__(self, lb_stats_dict: LbStatsDict):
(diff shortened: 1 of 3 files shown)


framework/rpc/grpc_testing.py
sergiitk, line 1: - [ ] [grpc/java/v1.62.x/branch/xds_k8s_lb](https://source.cloud.google.com/results/invocations/47bb48df-516d-4809-9c29-62dfcb7fb35e)
- [ ] [grpc/java/v1.62.x/branch/xds_url_map](https://source.cloud.google.com/results/invocations/8eab855b-f63e-4dcc-9fd4-a560c15b30ba)

PR #53, opened 2024-03-05
Fix naming convention for gamma resources

sergiitk:
1. Fix camelCase method names and variables
2. Move SSA resource names to class  constants
3. Explicit arguments in `_create_gamma_route` and `_create_*` SSA resources
4. Add missing `owner` label to SSA manifests

### DIFF
--- a/framework/test_app/runners/k8s/gamma_server_runner.py
+++ b/framework/test_app/runners/k8s/gamma_server_runner.py
@@ -243,23 +243,24 @@ class GammaServerRunner(KubernetesServerRunner):
 
         return servers
 
-    def create_session_affinity_policy(self, template: str):
+    def create_session_affinity_policy(self, template: str, **template_vars):
         self.session_affinity_policy = self._create_session_affinity_policy(
             template,
             session_affinity_policy_name=self.SESSION_AFFINITY_POLICY_NAME,
             namespace_name=self.k8s_namespace.name,
-            route_name=self.route_name,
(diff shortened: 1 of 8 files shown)


framework/test_app/runners/k8s/gamma_server_runner.py
sergiitk, line 1: - [ ] [grpc/core/v1.62.x/branch/linux/psm-csm](https://source.cloud.google.com/results/invocations/ceb7fdd7-0518-4340-a0f4-36553ddbec84)

framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -505,114 +505,159 @@ def _create_deployment(self, template, **kwargs) -> k8s.V1Deployment:
         )
         return deployment
 
-    def _create_gamma_route(self, template, **kwargs) -> k8s.GammaHttpRoute:
+    def _create_gamma_route(
+        self,
+        template: str,
```
XuanWang-Amos, line 510: Can we change template to kwargs?

XuanWang-Amos (approved):

framework/test_app/runners/k8s/k8s_base_runner.py
sergiitk: All kwargs become template vars. That's why template is expected to be a positional arg. This pattern is all across `_create_*` methods in this class.

PR #62, opened 2024-03-29
Kokoro shared lib: Switch CSM tests to the regional cluster

sergiitk:
Move CSM tests to a new regional cluster and support for regional cluster activation.
Note: the new cluster has a same name as the old one (which is allowed since it's a different cluster type): `psm-interop-csm`, region `us-central1`.

Ref b/298501683

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -53,11 +53,7 @@ activate_gke_cluster() {
       ;;
     GKE_CLUSTER_PSM_CSM)
       GKE_CLUSTER_NAME="psm-interop-csm"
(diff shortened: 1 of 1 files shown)


.kokoro/psm_interop_kokoro_lib.sh
sergiitk, line 1: - [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/79b77b94-d26e-46aa-b612-0efac359d293)

sergiitk: - [ ] [grpc/core/v1.62.x/branch/linux/psm-csm](https://source.cloud.google.com/results/invocations/08fe50b5-8649-4eb2-8806-597188f7683a)

PR #65, opened 2024-04-04
AppNet SSA test: add is_supported when CPP >= v1.62.x

sergiitk:
Add missing `is_supported` conditional to the AppNetSsaTest.

(plus some minor code style cleanup)

ref b/298501683
needed for cl/621686576

### DIFF
--- a/tests/app_net_ssa_test.py
+++ b/tests/app_net_ssa_test.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import List, Optional
+from typing import Final, List, Optional
 
 from absl import flags
 from absl.testing import absltest
+from typing_extensions import TypeAlias
 
 from framework import xds_k8s_testcase
+from framework.helpers import skips
 from framework.rpc import grpc_testing
 from framework.test_app import client_app
 from framework.test_cases import session_affinity_util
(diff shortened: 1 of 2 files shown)


sergiitk (commented):
- [ ] [grpc/core/v1.62.x/branch/linux/psm-csm](https://source.cloud.google.com/results/invocations/9f700157-e198-4445-bbbf-b5bb0b92c380)

tests/app_net_ssa_test.py
```diff
@@ -28,13 +30,20 @@
...
 
-# This is here temporarily to run this test separately from other app_net tests.
-# TODO(sergiitk): Move into app_net_test.py
+# Constants.
+_REPLICA_COUNT: Final[int] = 3
 
 
 class AppNetSsaTest(xds_k8s_testcase.AppNetXdsKubernetesTestCase):
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
```
yashykt, line 42: on which versions is this run right now?

yashykt (approved):

tests/app_net_ssa_test.py
sergiitk: Now it's not restricted, but prod buildscripts only exist for cpp v1.62.x+. However, if we want to enable it in staging, we need this loging in the test itself.

PR #33, opened 2024-02-07
[CSM O11y test] Ping GMP endpoint during test for debugging purpose

stanley-cheung:
- Ping GMP endpoint in client and server pod `<pod IP>:9464/metrics` during the CSM Observability test
- This will provide insight into what GMP sees from the OTel plugin, before sending metrics to Cloud Monitoring
- This will help debug in case we get these extra metrics time series without a valid data point, in a future test flake.

PR #72, opened 2024-04-18
Add App Net based CSM Observability Test

stanley-cheung:
- Added `app_net_csm_observability_test.py` that's largely a copy of `gamma/csm_observability_test.py` with some adjustments
- Refactor some flags to be passed to the k8s_xds_server_runner

This PR was based on top of #33.

Able to run both tests successfully on gamma-enabled cluster.


framework/test_app/runners/k8s/gamma_server_runner.py
```diff
@@ -103,13 +100,14 @@ def __init__(
             namespace_template=namespace_template,
             debug_use_port_forwarding=debug_use_port_forwarding,
             enable_workload_identity=enable_workload_identity,
+            enable_csm_observability=enable_csm_observability,
```
sergiitk, line 103: instead, let's put all this into `deployment_args`, see 
https://github.com/grpc/psm-interop/blob/2eea2ff909d58f95917c2f5515bdb705704fc22d/framework/test_app/runners/k8s/k8s_xds_server_runner.py#L34-L36

PR #76, opened 2024-04-30
helpers: save last test run log when using `./run.sh`

sergiitk:
A minor change, but helpful for sharing the log post-execution.
Now, when `run.sh` is used, the last run log is saved to `./out/psm-last-run.log`.


sergiitk:
cc @larry-safran

XuanWang-Amos (approved):

PR #79, opened 2024-05-03
Shared kokoro lib: address feedback

sergiitk:
Improves the transparency into the build scripts.
1. Reduces the number of the environment variables required to be set in per-lang buildscripts
2. Deprecate `kokoro_setup_test_driver` in favor of `psm::setup::kokoro`, more consistent with the local setup
3. Instead of implicitly sourcing `psm-interop-build-${GRPC_LANGUAGE}.sh`, have the per-lang buildscripts define `psm::lang::build_docker_images`


ejona86 (approved):
Thank you!

sergiitk:
@ejona86 Thank you for the good feedback!

PR #82, opened 2024-05-10
[AppNet] Replace NetworkServicesV1Alpha1 with NetworkServicesV1

sergiitk:
Replaces  `NetworkServicesV1Alpha1` with `NetworkServicesV1`.

Using `v1alpha1` version of network-services API to manage meshes is no longer needed. Service Routing Mesh features went GA and became available in `v1` network-services API in Q3 2022.

Notes:
1. This change only affects app_net (AKA Service Routing) tests, no other suites use network-services `v1alpha1`.
2. Security test suite uses network-services `v1beta1` (see `NetworkServicesV1Beta1` class). Security suite could probably be switched to `v1` too, but it might make sense to do this separately, and look into replacing  network-security `v1beta1` with `v1` as well.
3. After this change, Private API access key (set in `--private_api_key_secret_name`) will no longer be needed (unless we'll need to use private v1alpha1 APIs again.

### DIFF
--- a/framework/infrastructure/gcp/api.py
+++ b/framework/infrastructure/gcp/api.py
@@ -164,7 +164,7 @@ class GcpApiManager:
                 api_key=self.private_api_key,
                 visibility_labels=["NETWORKSERVICES_ALPHA"],
             )
-        elif version == "v1beta1":
+        elif version in ("v1", "v1beta1"):
             return self._build_from_discovery_v2(api_name, version)
 
         raise NotImplementedError(f"Network Services {version} not supported")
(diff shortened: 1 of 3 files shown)


sergiitk:
- [ ] [grpc/core/v1.63.x/branch/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/fc7107c5-8c95-48d2-9875-cb02f77839b1)
- [ ] [grpc/java/v1.63.x/branch/psm-security](https://source.cloud.google.com/results/invocations/a4ac02d3-6394-40e2-9941-4899cd747fa0)

sergiitk:
FYI @larry-safran @sourabhsinghs

gnossen (approved):

PR #88, opened 2024-05-23
Everything working for proof of concept

larry-safran:


larry-safran:
We have real running tests, don't need POC anymore.

PR #94, opened 2024-06-05
csm: Pass value to argument --enable_csm_observability

ejona86:
This is needed by the Java client/server, because there's a simplistic parsing implementation. Since we need the parsing to work cross-language, we always include a value. `--secure_mode` and `--fail_on_failed_rpcs` (unused by this framework, but defined in xds-test-descriptions.md) are both boolean flags that we pass a value.

### DIFF
--- a/kubernetes-manifests/client.deployment.yaml
+++ b/kubernetes-manifests/client.deployment.yaml
@@ -49,7 +49,7 @@ spec:
             % endif
             - "--print_response=${print_response}"
             % if enable_csm_observability:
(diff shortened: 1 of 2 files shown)


ejona86:
Go: https://source.cloud.google.com/results/invocations/a1abe2ae-a35e-4168-bff9-69ee852e2c2e
C++: 	
https://source.cloud.google.com/results/invocations/912987c6-bd27-4e43-8af7-7e5ea53d8570

sergiitk (approved):

PR #95, opened 2024-06-05
CSM O11y: Enable Java

ejona86:
Enable CSM Observability tests for Java >= 1.65.x

This is dependent on #94


sergiitk (approved):
Approved, as long as ad-hoc tests pass. Make sure to rebase it so the test isn't simply skipped.

PR #101, opened 2024-07-08
Add dualstack support to the framework behind flag

purnesh42H:


framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +628,54 @@ def create_target_proxy(self):
...
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME + "-v6")
         elif self.target_proxy:
             name = self.target_proxy.name
+        elif self.target_proxy_v6:
+             name = self.target_proxy_v6.name
         else:
             return
```
larry-safran, line 648: This should be wrapped in `if ipv6:`.  Otherwise, you are relying on the ipv4 being cleaned up before calling the ipv6 cleanup which add fragility.

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +628,54 @@ def create_target_proxy(self):
...
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME + "-v6")
         elif self.target_proxy and self.target_proxy_is_http:
             name = self.target_proxy.name
+        elif self.target_proxy_v6 and self.target_proxy_is_http:
+            name = self.target_proxy_v6.name
         else:
             return
```
larry-safran, line 670: Like in the case above, reduce fragility by wrapping in `if ipv6`.

framework/infrastructure/traffic_director.py
```diff
@@ -679,32 +719,52 @@ def find_unused_forwarding_rule_port(
...
+            if ipv6:
+                name = self.make_resource_name(self.FORWARDING_RULE_NAME)
+            else:
+                name = self.make_resource_name(self.FORWARDING_RULE_NAME + "-v6")
         elif self.forwarding_rule:
             name = self.forwarding_rule.name
+        elif self.forwarding_rule_v6:
+            name = self.forwarding_rule_v6.name
         else:
             return
```
larry-safran, line 760: wrap in `if ipv6` like the other cleanups.

kubernetes-manifests/server.deployment.yaml
```diff
@@ -42,6 +42,9 @@ spec:
           ${prestop_hook.main_lifecycle() if pre_stop_hook else ''}
           args:
             - "--port=${test_port}"
+            % if add_dualstack_support:
+            - "--address_type=IPV4"
+            % endif
```
larry-safran, line 47: This isn't needed, it is better to just leave it to the default which will use the wildcard like it does today.

purnesh42H: do you mean remove the if block and keep "--address_type=IPV4" as default?

larry-safran: No, I mean that we shouldn't specify `address_type` except on tests that specifically want to limit which IP version is used.  When the server sees no `address_type` argument it will use the wildcard address.  I made sure to specifically handle that case so that existing tests would not be affected.  (not relevant, but possibly interesting is that `IPV4_IPV6` is the default value used when no address_type is specified)

framework/infrastructure/gcp/compute.py
```diff
@@ -39,6 +39,7 @@ class ComputeV1(
     _WAIT_FOR_BACKEND_SLEEP_SEC = 4
     _WAIT_FOR_OPERATION_SEC = 60 * 10
     gfe_debug_header: Optional[str]
+    add_dualstack_support: Optional[bool]
```
sergiitk, line 42: nit: no need `Optional` for booleans, we can simply set it to False
```suggestion
    add_dualstack_support: bool = False
```

framework/infrastructure/traffic_director.py
```diff
@@ -593,8 +605,12 @@ def delete_alternative_url_map(self, force=False):
         self.compute.delete_url_map(name)
         self.url_map = None
 
-    def create_target_proxy(self):
-        name = self.make_resource_name(self.TARGET_PROXY_NAME)
+    def create_target_proxy(self, ipv6=False):    
+        if ipv6:
+            name = self.make_resource_name(self.TARGET_PROXY_NAME + "-v6")
```
sergiitk, line 610: feel free to add `TARGET_PROXY_NAME_IPV6` constant (similar to `FIREWALL_RULE_NAME_IPV6` we already have) to avoid doing `+ "-v6"`  all the time

framework/infrastructure/gcp/compute.py
purnesh42H: Done

framework/infrastructure/traffic_director.py
purnesh42H: Done

kubernetes-manifests/server.deployment.yaml
purnesh42H: Done

framework/infrastructure/traffic_director.py
purnesh42H: Done

framework/infrastructure/traffic_director.py
purnesh42H: Done

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +632,54 @@ def create_target_proxy(self):
...
+        else:
+            self.target_proxy = create_proxy_fn(name, self.url_map)
+
+    def delete_target_grpc_proxy(self, force=False, ipv6=False):
         if force:
-            name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
```
larry-safran, line 646: The `if` is backwards

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +632,54 @@ def create_target_proxy(self):
...
         self.target_proxy_is_http = False
 
-    def delete_target_http_proxy(self, force=False):
+    def delete_target_http_proxy(self, force=False, ipv6=False):
         if force:
-            name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
```
larry-safran, line 668: The `if` is backwards

framework/infrastructure/traffic_director.py
larry-safran: I don't see the change.

framework/infrastructure/traffic_director.py
larry-safran: I don't see the change.

framework/infrastructure/traffic_director.py
purnesh42H: Good catch. Done

framework/infrastructure/traffic_director.py
purnesh42H: Fixed

framework/infrastructure/traffic_director.py
purnesh42H: Pushed a new commit. It was overridden due to interchange between cloudtop and local

framework/infrastructure/traffic_director.py
purnesh42H: Pushed a new commit. It was overridden due to interchange between cloudtop and local

kubernetes-manifests/server.deployment.yaml
larry-safran: After refreshing I see the changes in the delete method, but still see the added if statement here that should be removed.

purnesh42H: Removed. Looks like that was overriden too

larry-safran (approved):

purnesh42H:
[x] Successfully Run `./run.sh bin/run_td_setup.py` without --add_dualstack_support flag
[x] Successfully Run `./run.sh bin/run_td_setup.py` --cmd=cleanup without --add_dualstack_support flag
[] Successfully Run `./run.sh bin/run_td_setup.py` with --add_dualstack_support flag
[] Successfully Run `./run.sh bin/run_td_setup.py` --cmd=cleanup with --add_dualstack_support flag

purnesh42H:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/ff3b08b1-8d16-4a47-8437-140289f95f75)

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +631,56 @@ def create_target_proxy(self):
...
+
+    def delete_target_grpc_proxy(self, force=False, ipv6=False):
         if force:
-            name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+        elif ipv6:
+            if self.target_proxy_v6:
```
XuanWang-Amos, line 647: Will there be a case it's `ipv6` but not `self.target_proxy_v6`? What will be the name?

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +631,56 @@ def create_target_proxy(self):
...
-    def delete_target_http_proxy(self, force=False):
+    def delete_target_http_proxy(self, force=False, ipv6=False):
         if force:
-            name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
+        elif ipv6:
+            if self.target_proxy_v6 and self.target_proxy_is_http:
```
XuanWang-Amos, line 670: Same here, if those are not true, what will be the `name`?

framework/infrastructure/traffic_director.py
```diff
@@ -679,32 +724,60 @@ def find_unused_forwarding_rule_port(
         # TODO(sergiitk): custom exception
         raise RuntimeError("Couldn't find unused forwarding rule port")
 
-    def create_forwarding_rule(self, src_port: int):
-        name = self.make_resource_name(self.FORWARDING_RULE_NAME)
+    def create_forwarding_rule(
+        self, src_port: int, ipv6=False, ip_address="0.0.0.0"
```
XuanWang-Amos, line 728: We should probably add type hints here since we have it for `src_port`.

framework/infrastructure/traffic_director.py
```diff
@@ -679,32 +724,60 @@ def find_unused_forwarding_rule_port(
...
+
         src_port = int(src_port)
         logging.info(
-            'Creating forwarding rule "%s" in network "%s": 0.0.0.0:%s -> %s',
+            'Creating forwarding rule "%s" in network "%s": %s:%s -> %s',
             name,
             self.network,
             src_port,
-            self.target_proxy.url,
+            ip_address,
```
XuanWang-Amos, line 743: `src_port` should be put after `ip_address`.

framework/infrastructure/traffic_director.py
```diff
@@ -679,32 +724,60 @@ def find_unused_forwarding_rule_port(
...
+    def delete_forwarding_rule(self, force=False, ipv6=False):
         if force:
-            name = self.make_resource_name(self.FORWARDING_RULE_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.FORWARDING_RULE_NAME_IPV6)
+            else:
+                name = self.make_resource_name(self.FORWARDING_RULE_NAME)
+        elif ipv6:
+            if self.forwarding_rule_v6:
+                name = self.forwarding_rule_v6.name
```
XuanWang-Amos, line 769: Ditto, should we consider what should be the `name` here in case `self.forwarding_rule_v6` is not true?

framework/xds_flags.py
```diff
@@ -190,6 +190,12 @@
     help="Whether to enable GFE debug headers and what value to use.",
 )
 
+ADD_DUALSTACK_SUPPORT = flags.DEFINE_bool(
+    "add_dualstack_support",
+    default=False,
+    help=(f"Add support for Dual Stack resources to the framework."),
```
XuanWang-Amos, line 196: [Minor] We don't need `f` for a fixed string.

framework/infrastructure/gcp/api.py
```diff
@@ -157,6 +157,8 @@ def compute(self, version: str):
             return self._build_from_discovery_v1(api_name, version)
         elif version == "v1alpha":
             return self._build_from_discovery_v1(api_name, "alpha")
+        elif version in ["v1beta", "v1beta1"]:
```
sergiitk, line 160: Curious why we have both `v1beta` and `v1beta1` here

framework/infrastructure/gcp/compute.py
```diff
@@ -61,10 +62,12 @@ def __init__(
         api_manager: gcp.api.GcpApiManager,
         project: str,
         gfe_debug_header: Optional[str] = None,
+        add_dualstack_support: bool = False,
         version: str = "v1",
     ):
         super().__init__(api_manager.compute(version), project)
         self.gfe_debug_header = gfe_debug_header
+        self.add_dualstack_support = add_dualstack_support
```
sergiitk, line 70: This is only used in `create_backend_service_traffic_director`. Let's make it a parameter to this method instead. No need for the object property.

framework/xds_flags.py
```diff
@@ -190,6 +190,12 @@
     help="Whether to enable GFE debug headers and what value to use.",
 )
 
+ADD_DUALSTACK_SUPPORT = flags.DEFINE_bool(
+    "add_dualstack_support",
```
sergiitk, line 194: Here and elsewhere: but let's name it `enable_dualstack`. 
Sorry, we should've suggested it much sooner, so now there's plenty of renaming todo.

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +631,56 @@ def create_target_proxy(self):
...
-        self.target_proxy = None
+        if ipv6:
+            self.target_proxy_v6 = None
+        else:
+            self.target_proxy = None
+
         self.target_proxy_is_http = False
 
-    def delete_target_http_proxy(self, force=False):
+    def delete_target_http_proxy(self, force=False, ipv6=False):
```
sergiitk, line 663: I don't see us ever calling this with ipv6=True. We definitely should clean it up.

framework/infrastructure/traffic_director.py
```diff
@@ -612,30 +631,56 @@ def create_target_proxy(self):
...
         if force:
-            name = self.make_resource_name(self.TARGET_PROXY_NAME)
+            if ipv6:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
+            else:
+                name = self.make_resource_name(self.TARGET_PROXY_NAME)
```
sergiitk, line 668: Branching in this method gets pretty confusing for no good reason. We should refactor this to be similar break it up similar to the firewall rules, see `delete_firewall_rules`, `delete_firewall_rule`, `delete_firewall_rule_ipv6`

framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -998,9 +998,12 @@ def _wait_pod_started(self, name, **kwargs) -> k8s.V1Pod:
...
         pod = self.k8s_namespace.get_pod(name)
-        logger.info(
-            "Pod %s ready, IP: %s", pod.metadata.name, pod.status.pod_ip
-        )
+
+        if hasattr(pod.status, "pod_ip_s"):  # if running with dualstack support
+            pod_ip_s = pod.status.pod_ip_s
+        else:
+            pod_ip_s = pod.status.pod_ip
+        logger.info("Pod %s ready, IP: %s", pod.metadata.name, pod_ip_s)
```
sergiitk, line 1006:
```suggestion
        if hasattr(pod.status, "pod_ip_s"):  # if running with dualstack support
            pod_ips = pod.status.pod_ip_s
        else:
            pod_ips = pod.status.pod_ip
        logger.info("Pod %s ready, IP: %s", pod.metadata.name, pod_ips)
```

It sucks that k8s calls it `pod_ip_s` - they generate APIs automatically. But no reason to keep their naming in  our local variables

sergiitk (changes requested):

framework/infrastructure/gcp/api.py
purnesh42H: This is from @larry-safran proof of concept https://github.com/grpc/psm-interop/pull/88/files

framework/infrastructure/traffic_director.py
purnesh42H: Refactored ipv6 creations and deletions to separate methods

framework/infrastructure/traffic_director.py
purnesh42H: Same, Refactored ipv6 creations and deletetions to separate methods so this case will be taken care now

framework/infrastructure/traffic_director.py
purnesh42H: Not present anymore after refactor

framework/infrastructure/traffic_director.py
purnesh42H: Should be fixed after refactor

framework/xds_flags.py
purnesh42H: Done

framework/infrastructure/gcp/compute.py
purnesh42H: Done

framework/xds_flags.py
purnesh42H: Done everywhere

framework/test_app/runners/k8s/k8s_base_runner.py
purnesh42H: Removed the local variable and directly logging the fields in both if and else block

framework/infrastructure/traffic_director.py
purnesh42H: I see. I was trying to avoid much code duplication but after combining dualstack it did look a bit complex. Refactored all creations and deletions to separate ipv6 functions.

framework/infrastructure/traffic_director.py
purnesh42H: Not present after the refactor

framework/infrastructure/traffic_director.py
```diff
@@ -693,18 +735,45 @@ def create_forwarding_rule(self, src_port: int):
             name, src_port, self.target_proxy, self.network_url
         )
         self.forwarding_rule = resource
+
+        return resource
+
+    def create_forwarding_rule_ipv6(self, src_port: int):
+        name = self.make_resource_name(self.FORWARDING_RULE_NAME_IPV6)
+        src_port = int(src_port)
```
XuanWang-Amos, line 743: [Minor] Do we still need converting `src_port` here if it was passed in as `int` already?

XuanWang-Amos (approved):

purnesh42H:
All test suites against 1.65.x after refactor

- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/c5d3e050-d0f2-45ce-907a-9da67457c294)
- [ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/9d190217-1d74-428e-bf40-dfbc59739fcf)
- [ ] [grpc/core/master/linux/grpc_xds_v3](https://source.cloud.google.com/results/invocations/bde10aa4-eec6-482d-9041-60929243bc61)
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/970f5c89-10af-4a18-9f10-9764c557d381)

framework/infrastructure/gcp/api.py
sergiitk: Spoke with Larry, let's make it `version == "v1beta"`

framework/infrastructure/traffic_director.py
```diff
@@ -129,11 +131,14 @@ def __init__(
...
         self.firewall_rule_ipv6: Optional[GcpResource] = None
         self.target_proxy: Optional[GcpResource] = None
+        self.target_proxy_ipv6: Optional[GcpResource] = None
         # TODO(sergiitk): remove this flag once target proxy resource loaded
         self.target_proxy_is_http: bool = False
         self.alternative_target_proxy: Optional[GcpResource] = None
         self.forwarding_rule: Optional[GcpResource] = None
+        self.forwarding_rule_ipv6: Optional[GcpResource] = None
         self.alternative_forwarding_rule: Optional[GcpResource] = None
+        self.enable_dualstack = xds_flags.ENABLE_DUALSTACK.value
```
sergiitk, line 141: 1. We need to move this to `framework.xds_k8s_testcase.XdsKubernetesBaseTestCase.setUpClass`, where the other flags are parsed.
2. Add `enable_dualstack` parameter to the `TrafficDirectorManager` constructor
3. Provide initiate `TrafficDirectorManager` with `enable_dualstack=self.enable_dualstack` in  `framework.xds_k8s_testcase.RegularXdsKubernetesTestCase`

framework/xds_flags.py
```diff
@@ -190,6 +190,12 @@
     help="Whether to enable GFE debug headers and what value to use.",
 )
 
+ENABLE_DUALSTACK = flags.DEFINE_bool(
+    "enable_dualstack",
+    default=False,
+    help=("Enable support for Dual Stack resources to the framework."),
```
sergiitk, line 196: Remove redundant parentheses  

```suggestion
    help="Enable support for Dual Stack resources to the framework.",
```

kubernetes-manifests/server.service.yaml
```diff
@@ -10,6 +10,12 @@ metadata:
     cloud.google.com/neg: '{"exposed_ports": {"${test_port}":{"name":"${neg_name}"}}}'
 spec:
   type: ClusterIP
+  % if enable_dualstack:
```
sergiitk, line 13: We need to set this variable KubernetesBaseRunner, see https://github.com/grpc/psm-interop/blob/fc57159059a70534fef4136a15659eca9e140869/framework/test_app/runners/k8s/k8s_xds_server_runner.py#L219-L226

kubernetes-manifests/client.deployment.yaml
```diff
@@ -66,6 +66,12 @@ spec:
               value: "true"
             - name: GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST
               value: "true"
+            % if enable_dualstack:
```
sergiitk, line 69: For this variable to be available in the template, we need to set it in `KubernetesClientRunner`. Add it to `ClientDeploymentArgs`, see https://github.com/grpc/psm-interop/blob/fc57159059a70534fef4136a15659eca9e140869/framework/test_app/runners/k8s/k8s_xds_client_runner.py#L29-L33

framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -998,9 +998,12 @@ def _wait_pod_started(self, name, **kwargs) -> k8s.V1Pod:
...
         pod = self.k8s_namespace.get_pod(name)
-        logger.info(
-            "Pod %s ready, IP: %s", pod.metadata.name, pod.status.pod_ip
-        )
+
+        if hasattr(pod.status, "pod_ip_s"):  # if running with dualstack support
+            pod_ips = pod.status.pod_ip_s
+        else:
+            pod_ips = pod.status.pod_ip
+        logger.info("Pod %s ready, IP: %s", pod.metadata.name, pod_ips)
```
sergiitk, line 1006: I see that we're not saving the ips anywhere, just logging. @larry-safran - are we going to need actual IPs anywhere? If yes, we should capture them in `_xds_test_client_for_pod`/`_xds_test_server_for_pod`, and save them to the new `ips` field in `XdsTestClient`/`XdsTestServer`

framework/infrastructure/traffic_director.py
```diff
@@ -171,13 +176,19 @@ def setup_routing_rule_map_for_grpc(self, service_host, service_port):
...
+
     def cleanup(self, *, force=False):
         # Cleanup in the reverse order of creation
         self.delete_firewall_rules(force=force)
         self.delete_forwarding_rule(force=force)
         self.delete_alternative_forwarding_rule(force=force)
         self.delete_target_http_proxy(force=force)
         self.delete_target_grpc_proxy(force=force)
+        if self.enable_dualstack:
+            self.delete_target_proxy_ipv6(force=force)
```
sergiitk, line 191: We need to clean up forwarding rule ipv6 too.

purnesh42H: Had both in delete_forwarding_rule but moved ipv6 to separate method to keep consistent with convention

kubernetes-manifests/client.deployment.yaml
purnesh42H: Done

kubernetes-manifests/server.service.yaml
purnesh42H: Done

framework/test_app/runners/k8s/k8s_base_runner.py
purnesh42H: Do you mean replacing existing ip field by ips and populate from kubernetes based on dualstack or not? Or you are suggesting to maintain separate field for ips and populate only in case of dualstack?

framework/xds_flags.py
purnesh42H: Done

framework/infrastructure/traffic_director.py
purnesh42H: Done

framework/test_app/runners/k8s/k8s_base_runner.py
larry-safran: We didn't need it before and it can always be retrieved from the `pod`.

framework/infrastructure/traffic_director.py
sergiitk: Forwarding rule should be deleted before target proxy, or it'll fail because of the dependency order

framework/test_app/runners/k8s/k8s_base_runner.py
sergiitk: Thanks

sergiitk (commented):
LGTM, but give it another local run against the dual-stack cluster to polish the issues like the one with incorrect deletion orders. Then resolve the merge conflict with master, and re-run the tests. I'll give it another look early tomorrow.

sergiitk:
Oh, and please save the output of the local run - I'd like to give it a check in case we missed anything

framework/infrastructure/gcp/api.py
purnesh42H: Done

purnesh42H:
All test suites against 1.65.x after rebase with latest main

- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/c68934d6-6df3-4aad-b452-f08619f016f7)
- [ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/88d2b23a-0c50-47f1-8b97-990496b008cd)
- [ ] [grpc/core/master/linux/grpc_xds_v3](https://source.cloud.google.com/results/invocations/4f05a519-ec26-43d7-8146-39422d482db4)
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/eba03e6a-41fe-45ed-80a5-3b99c366ef3f)

sergiitk:
PSM Security suite:
- [ ] [grpc/java/v1.65.x/branch/psm-security](https://source.cloud.google.com/results/invocations/8c8d58ab-0619-4f6a-8f05-daccb182b057)

sergiitk:
@purnesh42H noticed that the jobs you ran weren't using this framework branch. Note the jobs I started above:

```pbtxt
# [08:17:29 PDT] Parsed build request:
full_job_name: "grpc/java/v1.65.x/branch/psm-security"
scm_revision {
  github_scm_revision {
    commit_sha: "v1.65.x"
    name: "grpc-java"
    owner: "grpc"
    repository: "grpc-java"
  }
}
env_vars {
  key: "TEST_DRIVER_BRANCH"
  value: "add-dualstack-support"
}
env_vars {
  key: "TEST_DRIVER_REPO_OWNER"
  value: "purnesh42H"
}
```

Also - no need to run ` grpc/core/master/linux/grpc_xds_v3`, this is legacy framework, which isn't affected by any changes to this repo.

sergiitk:
Ah, my fault. For some reason I expected these env vars to show up in BUILD_CONFIG on this page: https://btx.cloud.google.com/invocations/88d2b23a-0c50-47f1-8b97-990496b008cd/details.

framework/xds_k8s_testcase.py
```diff
@@ -796,7 +799,11 @@ def setUp(self):
         self.client_namespace = KubernetesClientRunner.make_namespace_name(
             self.resource_prefix, self.resource_suffix
         )
-        self.client_runner = self.initKubernetesClientRunner()
+        self.client_runner = self.initKubernetesClientRunner(
+            deployment_args=ClientDeploymentArgs(
```
sergiitk, line 803: This is probably not a good idea. We don't really merge base deployment args when this method is overridden in the suite, f.e.

https://github.com/grpc/psm-interop/blob/fc57159059a70534fef4136a15659eca9e140869/tests/app_net_csm_observability_test.py#L190-L197

I think I have an idea have we can make it right. Let me help you with this part.

framework/xds_k8s_testcase.py
```diff
@@ -1068,6 +1079,7 @@ def initKubernetesServerRunner(self, **kwargs) -> KubernetesServerRunner:
             xds_server_uri=self.xds_server_uri,
             deployment_template="server-secure.deployment.yaml",
             debug_use_port_forwarding=self.debug_use_port_forwarding,
+            enable_dualstack=self.enable_dualstack,
```
sergiitk, line 1082: Now I wonder if we should've added it to `ServerDeploymentArgs` similar to the client. Let me help you with this too.

sergiitk (commented):
All looks good, except I suspect ClientDeploymentArgs won't work as we want to. Could you apply/accept suggestions I made, and I'll take a stab at getting the deployment args right.

framework/xds_k8s_testcase.py
sergiitk: Discussed IRL - let's take care of this in a follow-up PR.

framework/xds_k8s_testcase.py
sergiitk: Discussed IRL - let's take care of this in a follow-up PR.

framework/infrastructure/traffic_director.py
```diff
@@ -638,6 +669,19 @@ def delete_target_http_proxy(self, force=False):
...
+        if force:
+            name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
+        elif self.target_proxy_ipv6:
+            name = self.target_proxy_ipv6.name
+        else:
+            return
+        logger.info('Deleting Target HTTP proxy "%s"', name)
+        self.compute.delete_target_http_proxy(
+            name
+        )  # TODO: Support GRPC target proxy as well
```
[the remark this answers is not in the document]
sergiitk: I'll redo the suggestion

framework/infrastructure/traffic_director.py
```diff
@@ -638,6 +670,19 @@ def delete_target_http_proxy(self, force=False):
...
+        if force:
+            name = self.make_resource_name(self.TARGET_PROXY_NAME_IPV6)
+        elif self.target_proxy_ipv6:
+            name = self.target_proxy_ipv6.name
+        else:
+            return
+        logger.info('Deleting IPv6 Target HTTP proxy "%s"', name)
+        self.compute.delete_target_http_proxy(
+            name
+        )  # TODO: Support GRPC target proxy as well
```
sergiitk, line 683:
```suggestion
        # TODO: Delete Target GRPC Proxy when added in create_target_proxy_ipv6.
        logger.info('Deleting IPv6 Target HTTP proxy "%s"', name)
        self.compute.delete_target_http_proxy(name)
```

sergiitk (approved):
Two items to address in the follow-up:
1. Move server runner's enable_dualstack to `ServerDeploymentArgs`
2. Fix the issue with deployment_args not merged in suites that override `initKubernetesClientRunner`, see https://github.com/grpc/psm-interop/pull/101/files#r1688415338

PR #105, opened 2024-07-12
[URL Map] Fix TestBasicCsds lds assertion as per change in TD template

purnesh42H:
A recent change to the TD gRPC bootstrap genrator adds client_default_listerner_resource_name_template to the config. This template is used to determine the LDS resource name used for LDS requests.

Our test framework currently tries to do an exact match of the LDS resource name with the dial target [here](https://github.com/grpc/psm-interop/blob/657acc2637b34ea167c5289762f02be07dae7042/tests/url_map/csds_test.py#L61)

This is the expected value now `xdstp://traffic-director-global.xds.googleapis.com/envoy.config.listener.v3.Listener/<gcp_project_number>/<vpc_network_name/mesh_id>/<dial_target>`. So the target URI is still in the new LDS but with templating around it

We can fix the assertion as below

`self.assertIn(self.hostname(), xds_config.lds["name"])`

### DIFF
--- a/tests/url_map/csds_test.py
+++ b/tests/url_map/csds_test.py
@@ -58,7 +58,7 @@ class TestBasicCsds(xds_url_map_testcase.XdsUrlMapTestCase):
         )
         # Validate Listeners
         self.assertIsNotNone(xds_config.lds)
-        self.assertEqual(self.hostname(), xds_config.lds["name"])
+        self.assertIn(self.hostname(), xds_config.lds["name"])
         # Validate Route Configs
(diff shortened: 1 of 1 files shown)


arvindbr8 (approved):

purnesh42H:
[ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/caeb9292-a11d-4db6-a4c5-63742c3e97d4)

gnossen (approved):

PR #106, opened 2024-07-22
docs(README): intended relative links

guspan-tanadi:
Navigating example path links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ do this, navigate to this directory and run `./bin/freeze.sh`.
 
 There are many arguments to be passed into the test run. You can save the
 arguments to a config file ("flagfile") for your development environment.
-Use [`config/local-dev.cfg.example`](./config/local-dev.cfg.example)
+Use [`config/local-dev.cfg.example`](https://github.com/grpc/psm-interop/blob/main/config/local-dev.cfg.example)
 as a starting point:
 
 ```shell
@@ -268,7 +268,7 @@ Learn more about flagfiles in [abseil documentation](https://abseil.io/docs/pyth
 
 ## Test suites
 
-See the full list of available test suites in the [`tests/`](./tests) folder. 
+See the full list of available test suites in the [`tests/`](https://github.com/grpc/psm-interop/blob/main/tests) folder. 
 
 ### xDS Baseline Tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,7 +250,7 @@ do this, navigate to this directory and run `./bin/freeze.sh`.
 
 There are many arguments to be passed into the test run. You can save the
 arguments to a config file ("flagfile") for your development environment.
-Use [`config/local-dev.cfg.example`](https://github.com/grpc/grpc/blob/master/tools/run_tests/xds_k8s_test_driver/config/local-dev.cfg.example)
+Use [`config/local-dev.cfg.example`](./config/local-dev.cfg.example)
```
sergiitk, line 253:
```suggestion
Use [`config/local-dev.cfg.example`](https://github.com/grpc/psm-interop/blob/main/config/local-dev.cfg.example)
```

README.md
```diff
@@ -268,7 +268,7 @@ Learn more about flagfiles in [abseil documentation](https://abseil.io/docs/pyth
 
 ## Test suites
 
-See the full list of available test suites in the [`tests/`](https://github.com/grpc/grpc/tree/master/tools/run_tests/xds_k8s_test_driver/tests) folder. 
+See the full list of available test suites in the [`tests/`](./tests) folder.
```
sergiitk, line 271:
```suggestion
See the full list of available test suites in the [`tests/`](https://github.com/grpc/psm-interop/blob/main/tests) folder. 
```

README.md
```diff
@@ -304,7 +304,7 @@ python -m tests.security_test --flagfile="config/local-dev.cfg"
 ```
 
 ## Helper scripts
-You can use interop xds-k8s [`bin/`](https://github.com/grpc/grpc/tree/master/tools/run_tests/xds_k8s_test_driver/bin)
+You can use interop xds-k8s [`bin/`](./bin)
```
sergiitk, line 307:
```suggestion
You can use interop xds-k8s [`bin/`](https://github.com/grpc/psm-interop/blob/main/bin)
```

sergiitk (approved):
Thanks for noticing that the links point to the old location in grpc/grpc!
However, we should still preserve full URLs: GitHub doesn't handle relative links from README.md correctly. For example, try to click "Display the rich diff" and follow one of the links. Same goes for github-pages generated from the readme - relative links just break.

PR #111, opened 2024-07-25
Add DualStack tests to psm_interop_kokoro_lib.sh

aranjans:
Added dualstack test to  `psm_interop_kokoro_lib.sh`.

PR #112, opened 2024-07-25
Kokoro: fix dualstack test name

aranjans:
`test_dualstack` -> `dualstack_test`


sergiitk (approved):
Thanks for the fix - I've missed it in #111 review

PR #116, opened 2024-08-05
[fallback] Set xDS configuration from the client

eugeneo:
Instead of hardcoding resource values in the control plane implementation, have test runner set the resources for each test case.


docker/go-control-plane/fallback-control-plane.go
```diff
@@ -52,12 +48,11 @@ import (
...
-	nodeid   = flag.String("nodeid", "test-id", "Node ID")
-	upstream = flag.String("upstream", "localhost:3000", "upstream server")
+	port   = flag.Uint("port", 3333, "Port to listen on")
+	nodeid = flag.String("nodeid", "test-id", "Node ID")
 )
 
-type Filter struct {
+type resourceKey struct {
 	ResourceType string
 	ResourceName string
```
easwars, line 57: Missed in the last PR maybe. Any reason for these fields to be exported? If not, please unexport them.

protos/grpc/testing/xdsconfig/xdsconfig.proto
```diff
@@ -38,18 +38,29 @@ message StopOnRequestResponse {
...
-  optional string listener = 1;
-  string cluster = 2;
-  string upstream_host = 3;
-  uint32 upstream_port = 4;
-};
+// Request to set one or more resources.
+message SetResourcesRequest {
+  message ResourceToSet {
+    // Resource type
+    string type = 1;
```
easwars, line 45: Should this be type_url instead?

docker/go-control-plane/fallback-control-plane.go
```diff
@@ -86,45 +80,48 @@ func (srv *controlService) StopOnRequest(_ context.Context, req *xdsconfigpb.Sto
...
-	srv.version++
-	listener := controlplane.ListenerName
-	if req.Listener != nil {
-		listener = *req.Listener
+	if len(req.Resources) > 0 {
+		srv.version++
+	}
+	for _, resource := range req.Resources {
+		key := resourceKey{ResourceType: resource.Type, ResourceName: resource.Name}
+		contents := resource.Body
```
easwars, line 101: Optional:

```golang
if contents := resource.Body; contents != nil {
	body, err := contents.UnmarshalNew()
	if err != nil {
		log.Printf("Failed to parse %s/%s: %v", key.ResourceType, key.ResourceName, err)
		continue
	}
	srv.resources[key] = &body
} else {
	delete(srv.resources, key)
}
```

Or

```golang
for _, resource := range req.Resources {
	key := resourceKey{ResourceType: resource.Type, ResourceName: resource.Name}
	contents := resource.Body
	if contents == nil {
		delete(srv.resources, key)
		continue
	}
	body, err := contents.UnmarshalNew()
	if err != nil {
		log.Printf("Failed to parse %s/%s: %v", key.ResourceType, key.ResourceName, err)
		continue
	}
	srv.resources[key] = &body
}
```

I prefer the second approach. It feels more idiomatic Go where we avoid `else` branches wherever possible and the code is less indented.

eugeneo (commented):
Thanks for the reviews. I addressed the comments.

docker/go-control-plane/fallback-control-plane.go
easwars: Why do you feel that `continue` is too prone to errors? I can buy that for `goto` (or its equivalent in Go which is `break <Label>`.

docker/go-control-plane/README.md
```diff
@@ -14,3 +14,12 @@ the root of `grpc/psm-interop` checkout.
 ```
 docker build . -f docker/go-control-plane/Dockerfile
 ```
+
+## Local development
+
+Run the following command from this repository root to generate protocol files:
```
easwars, line 20: You are not generating protocol files here. I would say something like:
`Run the following command from this repository to compile the .proto files` or
`Run the following command from this repository to generate code from the .proto files`.

easwars (approved):
LGTM for the Go changes.

docker/go-control-plane/fallback-control-plane.go
eugeneo: It's personal preference. I feel like
```
if (error) {
   doBad();
} else {
   doGood();
}
```
is safer then
```
if (error) {
  doBad();
  continue;
}
doGood();
```
because missing the `continue` runs `doGood()` . It is quite similar to me to fallthrough in switch/case, that is well known to cause bugs.

easwars: This is like saying:

```
if (error) {
   doBad();
} else {
   doGood();
}
return
```

is safer than

```
if (error) {
   doBad();
   return
}
doGood();
return
```

eugeneo: As I said, this is a personal preference... Thanks for the review!

docker/go-control-plane/README.md
eugeneo: Done

tests/fallback_test.py
```diff
@@ -59,17 +75,28 @@ def start_client(self, port: int = None, name: str = None):
...
+    ):
         logger.debug('Starting control plane "%s"', name)
         return framework.helpers.docker.ControlPlane(
             self.process_manager,
             name=name,
             port=self.bootstrap.xds_config_server_port(index),
-            upstream=f"{_HOST_NAME.value}:{upstream_port}",
+            initial_resources=framework.helpers.xds_resources.build_listener_and_cluster(
+                listener_name=_LISTENER,
+                cluster_name=(
```
XuanWang-Amos, line 92: Can change to `cluster_name=cluster_name or f"initial_cluster_for_{name}"`

XuanWang-Amos (approved):

framework/helpers/docker.py
```diff
@@ -266,34 +280,28 @@ def __init__(
...
-            command=["--upstream", str(upstream), "--nodeid", manager.node_id],
+            command=["--nodeid", manager.node_id],
         )
+        self.initial_resources = initial_resources
+
+    def __enter__(self):
+        if not super().__enter__():
+            return None
+        self.update_resources(self.initial_resources)
+        return self
```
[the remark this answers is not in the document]
eugeneo: Done

tests/fallback_test.py
eugeneo: Done

eugeneo (commented):
Thank you for the review. I implemented the suggestions.

framework/helpers/docker.py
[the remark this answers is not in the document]
eugeneo: Reverted the change. pylint requires all code paths return values.

PR #120, opened 2024-08-13
Add a thread to recreate the IPv6 firewall rule every 30 seconds to counteract the enforcer deleting it.

larry-safran:
Short term fix to them not providing the exemption for the firewall rule.


tests/dualstack_test.py
```diff
@@ -128,6 +134,22 @@ def test_dualstack(self) -> None:
                 self.server_xds_host, self.server_xds_port
             )
 
+        with self.subTest("_start_firewall_rule_creation_thread"):
+
+            def recreate_firewall_rule():
+                time.sleep(30)
+                while self.firewall_rule_creation_should_stop is False:
```
gnossen, line 141: Please use [`threading.Event`](https://docs.python.org/3/library/threading.html) for inter-thread synchronization. The current approach may suffer from tearing.

tests/dualstack_test.py
```diff
@@ -128,6 +134,22 @@ def test_dualstack(self) -> None:
...
+                    self.td.create_firewall_rules(
+                        allowed_ports=self.firewall_allowed_ports,
+                        source_range=self.firewall_source_range,
+                        source_range_ipv6=self.firewall_source_range_ipv6,
+                    )
+                    time.sleep(30)
```
gnossen, line 147: Seems like this could be done with a single call to `time.sleep`

gnossen, line 147: Can the delay be pulled to a module-level constant? Also, 30 seems a bit slow to me. Maybe 10?

tests/dualstack_test.py
```diff
@@ -128,6 +134,22 @@ def test_dualstack(self) -> None:
                 self.server_xds_host, self.server_xds_port
             )
 
+        with self.subTest("_start_firewall_rule_creation_thread"):
+
+            def recreate_firewall_rule():
```
gnossen, line 139: Seems like this functionality is generally useful. Can we just incorporate it into the `--ensure_firewall` flag?

larry-safran:
No longer needed since the exemption is in place

PR #122, opened 2024-08-13
Add cluster for a fallback test

eugeneo:
Trying to address this Kokoro failure:
```
+ [21:10:41 UTC]	 Installing packages with apt, see install-apt.log
+ [21:11:21 UTC]	 Fetching GKE cluster credentials
ERROR: (gcloud.container.clusters.get-credentials) One of [--location, --zone, --region] must be supplied.
```

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -207,10 +207,10 @@ psm::dualstack::run_test() {
 #######################################
 # Fallback Test Suite setup.
 #
-# This test does not actively need GKE but Kokoro fails if no cluster is provided.
+# This test does not need GKE so setting variable to skip some init
 #######################################
 psm::fallback::setup() {
-  activate_gke_cluster GKE_CLUSTER_PSM_BASIC
+  NO_GKE_CLUSTER=1
 }
 
 #######################################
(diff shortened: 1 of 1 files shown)


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -206,9 +206,11 @@ psm::dualstack::run_test() {
 
 #######################################
 # Fallback Test Suite setup.
+#
+# This test does not actively need GKE but Kokoro fails if no cluster is provided.
 #######################################
 psm::fallback::setup() {
-  : # Placeholder for future setup commands
+  activate_gke_cluster GKE_CLUSTER_PSM_BASIC
```
XuanWang-Amos, line 213: I think this means we'll use `GKE_CLUSTER_PSM_BASIC` clusters for fallback tests? Should we create new clusters for fallback?

eugeneo: The test itself does not use GKE. This is only to satisfy the infra.

XuanWang-Amos: I'm not familiar with this part of the framework, if we don't need GKE clusters, do we have to use `activate_gke_cluster` here?

eugeneo: I do not know how to test this locally... I am now trying to bypass the cluster init. Please take another look.

XuanWang-Amos: Skip them makes more sense, LGTM.

XuanWang-Amos (approved):

PR #124, opened 2024-08-14
Authenticate with a Docker registry

eugeneo:
This looks like the last missing piece. I was able to run test to a completion on Kokoro with this patch.

[Successful Run](https://fusion2.corp.google.com/invocations/aeee37d0-a500-4b6f-bc3e-209765139bc9/log)


XuanWang-Amos (approved):
LGTM, can you attach the passed runs to description?

PR #125, opened 2024-08-15
Use channelz to determine the ADS connections state 

eugeneo:
This way we can detect when the client starts talking to an expected control plane. This makes the test more stable.

[Test run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fpsm-fallback/activity/9ddb32fc-899a-404b-9c17-70ee774d736b/log)

### DIFF
--- a/framework/helpers/docker.py
+++ b/framework/helpers/docker.py
@@ -268,9 +268,6 @@ class Client(GrpcProcess):
             command=[
                 "--server",
                 url,
-                # "--print_response",
-                # "true",
-                # "--verbose",
                 "--stats_port",
                 str(port),
             ],
@@ -310,6 +307,6 @@ class Client(GrpcProcess):
                     status = ch.data.state.state
                     break
             if status == expected_status:
-                return status
+                break
             time.sleep(poll_interval.microseconds * 0.000001)
         return status
--- a/framework/helpers/docker.py
+++ b/framework/helpers/docker.py
@@ -19,6 +19,7 @@ import math
 import pathlib
 import queue
 import threading
+import time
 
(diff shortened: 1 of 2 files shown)


framework/helpers/docker.py
```diff
@@ -318,8 +265,16 @@ def __init__(
...
             image=image,
             name=name,
-            command=[f"--server={url}", "--print_response"],
-            ports={DEFAULT_GRPC_CLIENT_PORT: port},
+            command=[
+                "--server",
+                url,
+                # "--print_response",
+                # "true",
+                # "--verbose",
```
XuanWang-Amos, line 273: Should we delete those?

framework/helpers/docker.py
```diff
@@ -333,7 +288,28 @@ def get_stats(self, num_rpcs: int):
...
+        channelz = ChannelzServiceClient(self.channel())
+        status = None
+        while datetime.datetime.now() < deadline:
+            status = None
+            for ch in channelz.list_channels():
+                if ch.data.target.endswith(str(port)):
+                    status = ch.data.state.state
+                    break
+            if status == expected_status:
+                return status
```
XuanWang-Amos, line 313: We're already checking status use `assertEqual`, should we consider just return a status here directly?

XuanWang-Amos (commented):
Please add a test run to description.

eugeneo (commented):
Thank you for the review. I added a link to a successful run (with your comments addressed) into the PR description.

XuanWang-Amos (approved):

PR #135, opened 2024-10-08
minor: Use native TypeAlias

sergiitk:
[TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) was added in 3.10, so now we don't need to import it from `typing_extensions` anymore. 

Updating it now everywhere so that the old-style import doesn't spread by-copypasting.


sergiitk:
Superseded by #136.

XuanWang-Amos (approved):

PR #143, opened 2024-12-03
unskip gamma_grpcroute test

arvindbr8:
The fix has already rolledout, we need to unskip this test. 

cc: @sergiitk


sergiitk (approved):
LGTM if test passes

PR #145, opened 2024-12-12
Move app-net tests from CSM to LB suite

shivaspeaks:
This is a small part of public preview post-launch clean up. 

Internal: b/333124034


XuanWang-Amos (approved):
Can you add why we're making this change to description?

gnossen (approved):

PR #153, opened 2025-02-17
Updates resource cleanup script to support gamma resources for Prod

AgraVator:


XuanWang-Amos:
Please link some test runs to make sure it does not break anything.

AgraVator:
> Please link some test runs to make sure it does not break anything.

Had triggered it from local, since the pipeline gets triggered from grpc/grpc which would have always pointed to psm-interop's main

XuanWang-Amos:
> > Please link some test runs to make sure it does not break anything.
> 
> Had triggered it from local, since the pipeline gets triggered from grpc/grpc which points to psm-interop's main unless I am mistaken.

If clean up is part of normal job, you can run an adhoc kokoro job and use `TEST_DRIVER` so the test will use your branch, you can find an example command in [go/grpc-blr-learnings](http://goto.google.com/grpc-blr-learnings) `Run adhoc kokoro job on prod` section.

AgraVator:
> > > Please link some test runs to make sure it does not break anything.
> > 
> > 
> > Had triggered it from local, since the pipeline gets triggered from grpc/grpc which points to psm-interop's main unless I am mistaken.
> 
> If clean up is part of normal job, you can run an adhoc kokoro job and use `TEST_DRIVER` so the test will use your branch, you can find an example command in [go/grpc-blr-learnings](http://goto.google.com/grpc-blr-learnings) `Run adhoc kokoro job on prod` section.

[Test run link]([[url](https://fusion2.corp.google.com/ci/guitar/projects/network_gke_staging_us-central1-a-grpc-psm-csm-tests/workflows/%2F%2Fcloud%2Fcluster%2Ftesting%2Fguitar%2Fnetwork:grpc-psm-interop-csm-gke-staging/activity/c99a1571-51a1-32e5-b5e9-5d5c465a5431/invocations/5ad660ed-4935-4b96-8034-222afc31fb20/targets/%2F%2Fcloud%2Fcluster%2Ftesting%2Fe2e_test%2Fnetwork%2Fproxyless_grpc:psm_interop_gamma_cpp_master_test/log)](url))

XuanWang-Amos (approved):

bin/cleanup/cleanup.py
```diff
@@ -413,6 +416,53 @@ def cleanup_client(
...
+    # Shorten the timeout to avoid waiting for the stuck namespaces.
+    # Normal ns deletion during the cleanup takes less two minutes.
+    ns.wait_for_namespace_deleted_timeout_sec = 5 * 60
+    server_runner = _GammaServerRunner(
+        k8s_namespace=ns,
+        deployment_name=deployment_name,
+        gcp_project=project,
+        network=network,
+        gcp_service_account=gcp_service_account,
+        gcp_api_manager=gcp_api_manager,
```
sergiitk, line 444: We need to pass `enable_workload_identity` from `xds_k8s_flags.ENABLE_WORKLOAD_IDENTITY.value`. This flag is true by default, so it's not a big deal it's omitted in the cleanup_server, but in gamma tests it's configured to be disabled by default in the flags.

sergiitk (approved):
Looks good, thank you. Noticed one thing that should speed up the process.

sergiitk:
@AgraVator Oh. And I've noticed the adhoc run link you shared earlier is for a different test, not a cleanup. However, I'm fine with merging it without the test run. The code seems fine, and the rollback should be easy if the cleanup isn't working.

PR #154, opened 2025-02-26
cloudrun: GKE client to cloudrun server test

eshitachandwani:
Tests the case where the client is in GKE and the server is running in cloudrun. Verifies that 100 RPCs reach the server correctly and return with the correct response.


bin/run_test_server_c6n.py
```diff
@@ -0,0 +1,66 @@
...
+# limitations under the License.
+"""
+Run test xds server.
+
+Typical usage examples:
+
+    # Help.
+    ./run.sh ./bin/run_test_server_cloud_run.py --help
+
+    # Run modes.
```
kannanjgithub, line 22: Removed these since we are not taking any args. Only have a single run mode without any arg.
Rename the file was well to have cloud_run instead of c6n.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2024 gRPC authors.
```
kannanjgithub, line 1: 2025

framework/infrastructure/gcp/compute.py
```diff
@@ -204,17 +209,22 @@ def backend_service_patch_backends(
...
+            backend_list = [{"group": backend.url} for backend in backends]
+        else:
+            if max_rate_per_endpoint is None:
+                max_rate_per_endpoint = 5
+
+            backend_list = [
+                {
+                    "group": backend.url,
+                    "balancingMode": "RATE",
+                    "maxRatePerEndpoint": max_rate_per_endpoint,
```
kannanjgithub, line 224: Have these 2 for cloud run backend as well.

framework/infrastructure/gcp/compute.py
```diff
@@ -651,6 +736,8 @@ def _execute(  # pylint: disable=arguments-differ
             )
             request.headers[DEBUG_HEADER_KEY] = self.gfe_debug_header
             request.add_response_callback(self._log_debug_header)
+        logger.info("Executing request: %s", request)
+        logger.info(request.to_json())
```
kannanjgithub, line 740: Remove this logging, it seems not really related to our changes.

framework/infrastructure/traffic_director_cloudrun.py
```diff
@@ -0,0 +1,119 @@
...
+Mesh = gcp.network_services.Mesh
+
+
+class TrafficDirectorCloudRunManager(TrafficDirectorAppNetManager):
+    MESH_NAME = "grpc-mesh"
+    SERVER_TLS_POLICY_NAME = "server-tls-policy"
+    CLIENT_TLS_POLICY_NAME = "client-tls-policy"
+    AUTHZ_POLICY_NAME = "authz-policy"
+    ENDPOINT_POLICY = "endpoint-policy"
+    CERTIFICATE_PROVIDER_INSTANCE = "google_cloud_private_spiffe"
```
kannanjgithub, line 41: These seem to be unused.

framework/infrastructure/traffic_director_cloudrun.py
```diff
@@ -0,0 +1,119 @@
...
+        # TODO(gnossen) PTAL at the pylint error
+        self.grpc_route: Optional[GrpcRoute] = None
+        self.http_route: Optional[HttpRoute] = None
+        self.mesh: Optional[Mesh] = None
+
+        # Managed resources
+        self.server_tls_policy: Optional[ServerTlsPolicy] = None
+        self.client_tls_policy: Optional[ClientTlsPolicy] = None
+        self.authz_policy: Optional[AuthorizationPolicy] = None
+        self.endpoint_policy: Optional[EndpointPolicy] = None
```
kannanjgithub, line 81: We won't need these policies for the cloud run CSM gRPC testcase, but only for the SPIFFE test.

framework/infrastructure/traffic_director_cloudrun.py
```diff
@@ -0,0 +1,119 @@
...
+                "capacityScaler": capacity_scaler,
+            }
+
+            if circuit_breakers is not None:
+                new_backend["circuitBreakers"] = circuit_breakers
+
+            new_backends.append(new_backend)
+
+        # self.backends.update(new_backends)
+        backend_service = self.backend_service
```
kannanjgithub, line 107: Directly use self.backend_service

framework/infrastructure/traffic_director_cloudrun.py
```diff
@@ -0,0 +1,89 @@
...
+        logging.info(
+            "Adding backends to Backend Service %s: %r",
+            self.backend_service.name,
+            new_backends,
+        )
+
+        self.compute.backend_service_patch_backends(
+            self.backend_service,
+            backends,
+            is_cloudrun=True,
```
AgraVator, line 88: Can we ensure that 'Cloud Run' is treated as two separate words consistently throughout?

bin/lib/common.py
```diff
@@ -59,6 +62,11 @@ def gcp_api_manager():
     return gcp.api.GcpApiManager()
 
 
+@functools.cache
+def c6n_api_manager():
```
AgraVator, line 66: Can we ensure that we are not using c6n anywhere ?

framework/xds_k8s_testcase.py
```diff
@@ -1395,3 +1400,177 @@ def debug_cert(cert):
...
+    def backend_service_add_serverless_neg_backends(self):
+        logger.info("Creating serverless NEG")
+        neg = self.compute_v1.create_serverless_neg(
+            self.server_namespace,
+            self.region,
+            self.server_namespace,
+        )
+        return neg
+
+    def assertXdsConfigExists(self, test_client: XdsTestClient):
```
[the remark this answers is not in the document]
eshitachandwani: What would be the benefit of doing that ? As far as I could see we overwrite the functions we can to change from the base class.

framework/xds_k8s_testcase.py
```diff
@@ -1395,3 +1400,177 @@ def debug_cert(cert):
             return "missing"
         sha1 = hashlib.sha1(cert)
         return f"sha1={sha1.hexdigest()}, len={len(cert)}"
+
+
+class CloudRunXdsKubernetesTestCase(RegularXdsKubernetesTestCase):
```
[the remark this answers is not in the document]
eshitachandwani: changed to make it extend SecurityXdsKubernetesTestCase , but I dont understand the part about using the same YAML , we are still using the same yaml as client.deployement.yaml

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
```diff
@@ -0,0 +1,141 @@
+# Copyright 2022 gRPC authors.
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,81 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+        Args:
+            name: The name of the NEG.
+            region: The region in which to create the NEG.
+            service_name: The name of the Cloud Run service.
+            service_name format is "namespaces/{namespace}/services/{service}"
+
+        Returns:
+            The NEG selfLink URL
+        """
+        name = name + "-neg"
```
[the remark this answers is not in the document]
XuanWang-Amos: I mean if we call this method like this: `create_serverless_neg(name="emchandwani-psm-mesh", ...)`, the name of the neg will actually be `emchandwani-psm-mesh-neg`, what I'm suggestion is to use the name as it is, if we want the name to end with `-neg`, then just call this function like this: `create_serverless_neg(name="emchandwani-psm-mesh-neg", ...)`.

framework/xds_k8s_testcase.py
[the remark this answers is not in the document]
kannanjgithub: To avoid repetition of logic. If the assertion for Xds config for GKE case and cloud run case was vastly different, it would make more sense to override it completely but here the only difference is that in the GKE case we assert for endpoint_config but not in the case of cloud run Xds config. 
There may be an argument that the base class is polluted with subclass logic, and it is valid. For that I have this new suggestion: let the base class _assertXdsConfig_ only validate LDS, RDS and CDS, and then call another method mayBeAssertEdsConfig(seen) . Base class method will do the EDS config assertions but the cloud run subclass override of mayBeAssertEdsConfig will just be a no-op  (_pass_ statement).

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,102 @@
...
+        service_name: str,
+        image_name: str,
+        *,
+        test_port: int = DEFAULT_TEST_PORT,
+    ):
+        if not service_name:
+            raise ValueError("service_name cannot be empty or None")
+        if not image_name:
+            raise ValueError("image_name cannot be empty or None")
+        service_name = service_name[:49]
```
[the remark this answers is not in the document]
kannanjgithub: Can you figure this out by printing what is that first 48 characters being extracted? If there is some delimiter at index 48 we can use that to split.

framework/infrastructure/gcp/compute.py
```diff
@@ -600,19 +600,19 @@ def create_serverless_neg(self, name: str, region: str, service_name: str):
...
             name: The name of the NEG to delete.
             zone: The zone of the NEG.
         """
         try:
-            logger.info("Deleting serverless NEG %s in %s", name, zone)
+            logger.info("Deleting serverless NEG %s in %s", name, region)
             operation = (
-                self.api.networkEndpointGroups()
+                self.api.regionNetworkEndpointGroups()
                 .delete(
```
kannanjgithub, line 614: This change looks like impacting existing tests that use zonal deployment for NEGs.

framework/infrastructure/gcp/compute.py
```diff
@@ -635,12 +635,15 @@ def delete_serverless_neg(self, name: str, zone: str):
...
-        neg = (
+        try:
+            neg = (
             self.api.regionNetworkEndpointGroups()
             .get(project=self.project, networkEndpointGroup=name, region=region)
             .execute()
-        )
-        # TODO(sergiitk): dataclass
+            )
+        except Exception as e:
```
kannanjgithub, line 650: This  try catch only creates redundant log as the caller site already has it.

kubernetes-manifests/client.deployment.yaml
```diff
@@ -41,6 +41,9 @@ spec:
             - "--qps=${qps}"
             - "--rpc=${rpc}"
             - "--metadata=${metadata}"
+            % if secure_mode:
+            - "--secure_mode=${secure_mode}"
```
kannanjgithub, line 45: We don't need this change. --secure_mode=false/true is always passed here https://github.com/grpc/psm-interop/blob/acf694ea439d734e78a65d551fdee7802e6bd24e/framework/test_app/runners/k8s/k8s_xds_client_runner.py#L190

kannanjgithub (changes requested):
Left review comments.

framework/xds_k8s_testcase.py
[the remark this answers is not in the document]
kannanjgithub: Yes, you can ignore the part of my comment where I wrongly thought there was an additional yaml created for cloud run.

framework/infrastructure/gcp/cloud_run.py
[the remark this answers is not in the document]
eshitachandwani: It doesn't make a difference, I checked the complete service name is less than 49 characters itself. I have removed it. Not sure why it was there .

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
eshitachandwani: done.

framework/infrastructure/gcp/compute.py
eshitachandwani: Right! Changed.

kubernetes-manifests/client.deployment.yaml
eshitachandwani: Right!

framework/xds_k8s_testcase.py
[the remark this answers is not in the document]
eshitachandwani: Done.

kubernetes-manifests/client.deployment.yaml
kannanjgithub: I wrongly thought the args will be dynamically become deployment parameters. 
Can we use client-secure.deployment.yaml instead of client-deployment.yaml for CloudRun?

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -55,50 +88,32 @@ def deploy_service(
...
+             "template":
+                       {
+                            "containers": [
+                               {
+                                  "image": image_name,
+                                    "ports": [{"containerPort": test_port, "name": "h2c"}],            
+                                }
+                             ],
+                        },
+                    }
```
kannanjgithub: nit: Alignment does not look correct.

kannanjgithub (approved):

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,138 @@
...
+        *,
+        test_port: int = DEFAULT_TEST_PORT,
+    ):
+        if not service_name:
+            raise ValueError("service_name cannot be empty or None")
+        if not image_name:
+            raise ValueError("image_name cannot be empty or None")
+
+        try:
+            service_body = {}
```
XuanWang-Amos, line 109: [Nit] We don't need this line.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,138 @@
...
+                            ],
+                        }
+                    ],
+                },
+            }
+
+            self.create_cloud_run_resource(
+                self.service, service_name, service_body
+            )
+            logger.info("Deploying Cloud Run service '%s'", service_name)
```
XuanWang-Amos, line 127: We should log this before calling `create_cloud_run_resource`?

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,138 @@
...
+            return self.get_service_uri(service_name)
+
+        except Exception as e:  # noqa pylint: disable=broad-except
+            logger.exception("Error deploying Cloud Run service: %s", e)
```
XuanWang-Amos, line 131: Don't we need `raise` after log?

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
XuanWang-Amos: Looks like the name is still `name = name + "-neg"`?

framework/infrastructure/traffic_director_cloud_run.py
```diff
@@ -0,0 +1,89 @@
...
+                "maxRatePerEndpoint": max_rate_per_endpoint,
+                "capacityScaler": capacity_scaler,
+            }
+
+            if circuit_breakers is not None:
+                new_backend["circuitBreakers"] = circuit_breakers
+
+            new_backends.append(new_backend)
+
+        # backend_service = self.backend_service
```
XuanWang-Amos, line 77: [Nit] Remove unused line.

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
```diff
@@ -0,0 +1,141 @@
...
+    """Runs xDS Test Client/Server on Cloud Run."""
+
+    project: str
+    service_name: str
+    image_name: str
+    network: Optional[str] = None
+    tag: str = "latest"
+    region: str = "us-central1"
+    current_revision: Optional[str] = None
+    # gcp_project: Optional[str] = None
```
XuanWang-Amos, line 55: [Nit] Remove unused line.

framework/infrastructure/gcp/cloud_run.py
eshitachandwani: Done.

framework/infrastructure/gcp/cloud_run.py
eshitachandwani: Correct done.

framework/infrastructure/gcp/cloud_run.py
eshitachandwani: Right!

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
eshitachandwani: right! I changed the name passed to the function and missed to remove the `-neg` from here.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,81 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+            "name": name,
+            "networkEndpointType": "SERVERLESS",
+            "cloudRun": {"service": service_name},
+        }
+
+        try:
+            logger.info("Creating serverless NEG %s in %s", name, region)
+            self.api.regionNetworkEndpointGroups().insert(
+                project=self.project, region=region, body=neg_body
+            ).execute()
```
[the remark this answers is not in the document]
eshitachandwani: It keeps waiting indefinately. It is working correctly without the _wait.

kubernetes-manifests/client.deployment.yaml
eshitachandwani: what would be the benefit of doing that ? in the client.deployment.yaml also we have a check for the secure mode to be set only when specified. and the other changes will need to be done to the client-secure.deployment.yaml also.

XuanWang-Amos (approved):
Change LGTM, please add a test run to description before merge.

sergiitk:
@eshitachandwani - let me give it a final look today before merging this

config/common.cfg
```diff
@@ -1,5 +1,5 @@
 --resource_prefix=psm-interop
---td_bootstrap_image=us-docker.pkg.dev/grpc-testing/trafficdirector/td-grpc-bootstrap:ad35743ff01c0dc16a785e421591d7d52f3cdda4
+--td_bootstrap_image=us-docker.pkg.dev/grpc-testing/trafficdirector/td-grpc-bootstrap:e4622a2cb6956f85da18eb0aa9daec839c2e38b5
```
sergiitk, line 2: Please make this change is a separate PR. Staging should be updated too. cc @arvindbr8

eshitachandwani: Removed it from here , will put it in different PR.

eshitachandwani:
@XuanWang-Amos  Added a kokoro run.

XuanWang-Amos (approved):

tests/cloud_run_csm_inbound_test.py
```diff
@@ -12,37 +12,58 @@
...
 logger = logging.getLogger(__name__)
-flags.adopt_module_key_flags(xds_k8s_testcase)
 
+_XdsTestServer = xds_k8s_testcase.XdsTestServer
+_XdsTestClient = xds_k8s_testcase.XdsTestClient
 
-class CloudRunCsmInboundTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
+
+class CloudRunCsmInboundTest(xds_k8s_testcase.CloudRunXdsKubernetesTestCase):
     def test_gke_to_cloud_run_setup(self):
```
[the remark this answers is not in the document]
eshitachandwani: This is what abhishek has in guitar workflow already!

config/common.cfg
```diff
@@ -1,5 +1,5 @@
 --resource_prefix=psm-interop
---td_bootstrap_image=us-docker.pkg.dev/grpc-testing/trafficdirector/td-grpc-bootstrap:e4622a2cb6956f85da18eb0aa9daec839c2e38b5
+--td_bootstrap_image=us-docker.pkg.dev/grpc-testing/trafficdirector/td-grpc-bootstrap:ad35743ff01c0dc16a785e421591d7d52f3cdda4
```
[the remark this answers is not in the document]
eshitachandwani: I think you are looking at some commit where I have reverted the change , I dont see it in files changed. `ad35743ff01c0dc16a785e421591d7d52f3cdda4` this is the old version. I have changed to a new version in different PR.

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -363,25 +366,6 @@ psm::csm::get_tests() {
...
-# Executes CSM test case
-# Globals:
-#   PSM_TEST_FLAGS: The array with flags for the test
-# Arguments:
-#   Test case name
-# Outputs:
-#   Writes the output of test execution to stdout, stderr
-#   Test xUnit report to ${TEST_XML_OUTPUT_DIR}/${test_name}/sponge_log.xml
-#######################################
-psm::csm::run_test() {
```
sergiitk, line 376: Why is this removed?

tests/cloud_run_csm_inbound_test.py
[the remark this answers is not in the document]
sergiitk: Agree with Xuan, let's rename it. Test name here takes priority. We should rename it in staging too.

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
sergiitk: Please verify why it waits indefinitely. Is the condition wrong?

sergiitk (changes requested):
High-level comments so far:
1. Please resolve completed conversations. Without it, we won't be able to merge the PR. I've resolved a few that were clear to me, please do the rest.
2. Merge the `main` branch to the pick up latest bootstrap version we meged earlier.
3. Because common files are changed, we need to run all test suites to confirm they're not broken.

I'm continuing reviewing the PR in more details.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+import abc
+import logging
+
+from googleapiclient import discovery
+
+from framework.infrastructure import gcp
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TEST_PORT = 8080
```
sergiitk, line 23: nit
```suggestion
DEFAULT_TEST_PORT: Final[int] = 8080
```

You'll need to do `from typing import Final` too

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+DEFAULT_TIMEOUT = 600
```
sergiitk, line 24: This is unused, remove

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class CloudRunApiManager(
+    gcp.api.GcpStandardCloudApiResource, metaclass=abc.ABCMeta
```
sergiitk, line 28: This is a new-style API, it should be extending `GcpProjectApiResource`. I think this may fix the logic of waiting on the delete operation too.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @property
+    def api_version(self) -> str:
+        """Returns the API version for Cloud Run."""
+        return "v2"
+
+    def create_cloud_run_resource(
+        self, service: discovery.Resource, service_name: str, body: dict
+    ):
+        service.projects().locations().services().create(
```
sergiitk, line 64: the new-style api utilizes `GcpStandardCloudApiResource._create_resource` - should simplify things a bit

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+    def api_name(self) -> str:
+        """Returns the API name for Cloud Run."""
+        return "run"
+
+    @property
+    def api_version(self) -> str:
+        """Returns the API version for Cloud Run."""
+        return "v2"
+
+    def create_cloud_run_resource(
```
sergiitk, line 61: You're creating service here, right? DId you mean `create_cloud_run_service`? 
In fact, `CloudRunApiManager` already indicates it's cloud run. We should simplify this to `create_service`, and rename `get_` and `_delete` methods accordingly

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+        self, service: discovery.Resource, service_name: str, body: dict
+    ):
+        service.projects().locations().services().create(
+            parent=self._parent, serviceId=service_name, body=body
+        ).execute()
+
+    def get_cloud_run_resource(
+        self, service: discovery.Resource, service_name: str
+    ):
```
sergiitk, line 70: Instead, return a dataclass object. See `framework.infrastructure.gcp.network_services.NetworkServicesV1.get_mesh` and `framework.infrastructure.gcp.network_services.Mesh` for example.

Feel free to only parse fields that you need.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+
+    def get_service_uri(self, service_name: str) -> str:
+        response = self.get_cloud_run_resource(self.service, service_name)
+        return response.get("uri")
+
+    def delete_cloud_run_resource(
+        self, service: discovery.Resource, service_name: str
+    ):
+        service.projects().locations().services().delete(
```
sergiitk, line 85: similar to create, we should use `framework.infrastructure.gcp.api.GcpStandardCloudApiResource._delete_resource`

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+            name=self.resource_full_name(service_name, "services", self.region)
+        ).execute()
+
+    def deploy_service(
```
sergiitk, line 89: This a business logic not tied to wrapping API calls. It should be in an ApiManager, which controls "what" resources we create, while this library controls "how" we talk to the API itself. Move this method to `TrafficDirectorCloudRunManager`.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,131 @@
...
+        except Exception as e:  # noqa pylint: disable=broad-except
+            logger.exception("Error deploying Cloud Run service: %s", e)
+            raise
+
+    def delete_service(self, service_name: str):
+        try:
+            self.delete_cloud_run_resource(self.service, service_name)
+        except Exception as e:
+            logger.exception("Error deleting service: %s", e)
+            raise
```
sergiitk, line 131: This method is redundant, it just calls another method, which we'll rename to `delete_service` anyway.

sergiitk (changes requested):
framework/infrastructure/gcp/cloud_run.py feedback

framework/infrastructure/gcp/cloud_run.py
sergiitk: Ah, no, I'm wrong here. `GcpStandardCloudApiResource` is the right class here.

framework/infrastructure/gcp/api.py
```diff
@@ -520,19 +530,25 @@ def parent(self, location: Optional[str] = GLOBAL_LOCATION):
             location = self.GLOBAL_LOCATION
         return f"projects/{self.project}/locations/{location}"
 
-    def resource_full_name(self, name, collection_name):
-        return f"{self.parent()}/{collection_name}/{name}"
+    def resource_full_name(
+        self, name, collection_name, location: Optional[str] = None
```
sergiitk, line 534: Instead of `Optional[str] = None`, assign the default value to `GLOBAL_LOCATION`:
```suggestion
        self, name, collection_name, location: str = GLOBAL_LOCATION
```

framework/infrastructure/gcp/api.py
```diff
@@ -190,6 +190,16 @@ def networksecurity(self, version):
...
         raise NotImplementedError(f"Network Security {version} not supported")
 
+    @functools.lru_cache(None)
+    def cloudrun(self, version):
+        api_name = "run"
+        if version == "v2":
+            return self._build_from_discovery_v2(
+                api_name,
+                version,
+            )
```
sergiitk, line 200: nit:
```suggestion
            return self._build_from_discovery_v2(api_name, version)
```

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,78 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+        """
+        neg_body = {
+            "name": name,
+            "networkEndpointType": "SERVERLESS",
+            "cloudRun": {"service": service_name},
+        }
+
+        try:
+            logger.info("Creating serverless NEG %s in %s", name, region)
+            self.api.regionNetworkEndpointGroups().insert(
```
sergiitk, line 591: This should use `self._insert_resource`, which configures library-level parameters (f.e. `num_retries`), and handles errors correctly. . See any other `create_` method in this class for example.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,78 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+                project=self.project, region=region, body=neg_body
+            ).execute()
+            neg = self.get_serverless_network_endpoint_group(name, region)
+            logger.info("Created serverless neg %s ", neg)
+            return neg
+
+        except Exception as e:
```
sergiitk, line 598: There's almost never a good reason to catch bare exceptions. Not that `Exception` masks errors like `TypeError` which is often is code typo, literally.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,78 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+    def delete_serverless_neg(self, name: str, region: str):
+        """Deletes a serverless NEG.
+
+        Args:
+            name: The name of the NEG to delete.
+            zone: The zone of the NEG.
+        """
+        try:
+            logger.info("Deleting serverless NEG %s in %s", name, region)
+            self.api.regionNetworkEndpointGroups().delete(
```
sergiitk, line 611: Use `_delete_resource`, similar to all other `delete_` methods

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,78 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+                return
+            logger.exception("Error deleting serverless NEG: %s", error)
+            raise
+        except Exception as e:
+            logger.exception("Error deleting serverless NEG: %s", e)
+            raise
+
+    def get_serverless_network_endpoint_group(self, name, region):
+        neg = (
+            self.api.regionNetworkEndpointGroups()
```
sergiitk, line 633: same - use `self._get_resource`

sergiitk (changes requested):
framework/infrastructure/gcp/compute.py feedback

framework/infrastructure/gcp/compute.py
eshitachandwani: Removed.

bin/run_test_server_cloud_run.py
```diff
@@ -0,0 +1,51 @@
...
+from framework import xds_k8s_flags
+
+logger = logging.getLogger(__name__)
+
+flags.adopt_module_key_flags(xds_flags)
+flags.adopt_module_key_flags(xds_k8s_flags)
+flags.adopt_module_key_flags(common)
+
+
+def main(argv):
```
kannanjgithub, line 39: Is this script actually functional? Since we are not actually using it in the tests, we could remove it unless it is actually working.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +568,78 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+            service_name: The name of the Cloud Run service.
+            service_name format is "namespaces/{namespace}/services/{service}"
+
+        Returns:
+            The NEG selfLink URL
+        """
+        neg_body = {
+            "name": name,
+            "networkEndpointType": "SERVERLESS",
+            "cloudRun": {"service": service_name},
```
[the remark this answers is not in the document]
kannanjgithub: nit: Update the comment to 
"Note: Cloud Run is not the only service that uses serverless NEGs but we are only supporting Cloud Run right now."
and move it to about the "cloudRun" key.

framework/infrastructure/gcp/cloud_run.py
eshitachandwani: Done.

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
eshitachandwani: Done.

bin/run_test_server_cloud_run.py
eshitachandwani: yes, it is functional.

framework/infrastructure/gcp/cloud_run.py
eshitachandwani: Not really needed , but it becomes easier if we ever want to reuse it and well defined. I am not sure of the right way , so I can do it either way.

eshitachandwani: > the new-style api utilizes `GcpStandardCloudApiResource._create_resource` - should simplify things a bit

Simply reusing `GcpStandardCloudApiResource._create_resource` had some issues, becuase the `_wait` function used in it waits for the `done` keyword in result of the operation , but for cloudrun API , the response doesnt incluse `done` keyword. It is also mentioned [here](https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.operations#Operation) . So , I have added a argument for cloudrun and checked for the condition status. Not sure if this is the right way or if we should use something more generic.

eshitachandwani:
updated the kokoro run with latest changes.

framework/infrastructure/gcp/cloud_run.py
sergiitk: Apparently it does include it, but only after the operation is done. Was an easy fix: #166.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class CloudRunApiManager(
+    gcp.api.GcpStandardCloudApiResource, metaclass=abc.ABCMeta
+):
+    project: str
+    region: str
+    _parent: str
+    service: discovery.Resource
+    api_manager: gcp.api.GcpApiManager
```
sergiitk, line 52: all these are unused outside of the __init__ method, please convert to local variables

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, project: str, region: str):
+        if not project:
+            raise ValueError("Project ID cannot be empty or None.")
+        if not region:
+            raise ValueError("Region cannot be empty or None.")
+        self.api_manager = gcp.api.GcpApiManager(v2_discovery_uri=DISCOVERY_URI)
+        self.project = project
+        self.region = region
+        service: discovery.Resource = self.api_manager.cloudrun("v2")
+        self.service = service
```
sergiitk, line 63: No need to save it as a object property. Take a look at `_NetworkSecurityBase` - it should be enough to just do

```
super().__init__(api_manager.networksecurity(self.api_version), project)
```

and then use `self.api` instead of `self.service`

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+        self._parent = f"projects/{self.project}/locations/{self.region}"
+        super().__init__(self.service, project)
+
+    @property
+    def api_name(self) -> str:
+        """Returns the API name for Cloud Run."""
+        return "run"
```
sergiitk, line 70: So apparently this is only used for logs - I checked. Now it produces logs like this:
```
I0408 15:16:42.034237 140704314641152 api.py:543] Creating run resource:
I0408 15:16:42.709327 140704314641152 api.py:607] Waiting 600 sec for run operation id: projects/sergiitk-grpc-gke/locations/us-central1/operations/98f15b71-a6a6-40c3-ab07-5ce29c3770a0
```

Which is a bit confusing, especially `Waiting 600 sec for run operation id`. Let's make it `cloud_run` or `Cloud Run` for clarity.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def api_version(self) -> str:
+        """Returns the API version for Cloud Run."""
+        return "v2"
+
+    def create_service(
+        self, service: discovery.Resource, service_name: str, body: dict
```
sergiitk, line 78: no need to pass `service`, simply use `self.api`.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+    ) -> GcpResource:
+        return self._create_resource(
+            collection=service.projects().locations().services(),
```
sergiitk, line 81: Because `self.api.projects().locations()` is used everywhere, I create a shortcut for it like so: https://github.com/grpc/psm-interop/blob/12922820b1d662b8a1b7a11d9199c4c606a49923/framework/infrastructure/gcp/network_security.py#L111

And then call `self._api_locations.servicename()`, like so: https://github.com/grpc/psm-interop/blob/12922820b1d662b8a1b7a11d9199c4c606a49923/framework/infrastructure/gcp/network_security.py#L159

Also because you only use `self.api.projects().locations().services()` you can even store it as `self._services_collection`

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +567,60 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+        }
+        logger.info("Creating serverless NEG %s in %s", name, region)
+        self._insert_resource(
+            self.api.regionNetworkEndpointGroups(),
+            neg_body,
+            region=region,
+        )
+
+        neg = self.get_serverless_network_endpoint_group(name, region)
+        logger.info("Created serverless neg %s ", neg)
```
sergiitk, line 598: be consistent in spelling, elsewhere it's called NEG
```suggestion
        logger.info("Created serverless NEG %s ", neg)
```

framework/infrastructure/gcp/compute.py
```diff
@@ -579,12 +642,25 @@ def _exists_resource(
         return "items" in resp and resp["items"]
 
     def _insert_resource(
-        self, collection: discovery.Resource, body: Dict[str, Any]
+        self,
+        collection: discovery.Resource,
+        body: Dict[str, Any],
```
sergiitk, line 647: nit: use `dict[]` instead of deprecated `Dict[]`

framework/infrastructure/gcp/compute.py
```diff
@@ -652,11 +735,14 @@ def _execute(  # pylint: disable=arguments-differ
             request.headers[DEBUG_HEADER_KEY] = self.gfe_debug_header
             request.add_response_callback(self._log_debug_header)
         operation = request.execute(num_retries=self._GCP_API_RETRIES)
-        logger.debug("Operation %s", operation)
-        return self._wait(operation["name"], timeout_sec)
+        logger.info("Operation %s", operation)
```
sergiitk, line 738: Please revert to `debug`.  You can enable debug logs with 

```
--logger_levels=__main__:DEBUG,framework.infrastructure.gcp:DEBUG
```

or more global

```
--logger_levels=__main__:DEBUG,framework:DEBUG
```

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+                                {"containerPort": test_port, "name": "h2c"}
+                            ],
+                        }
+                    ],
+                },
+            }
+
+            logger.info("Deploying Cloud Run service '%s'", service_name)
+            self.create_service(self.service, service_name, service_body)
+            return self.get_service_uri(service_name)
```
sergiitk, line 143: instead, return full CloudRun and then just access result.uri

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+
+# Type aliases
+GcpResource = gcp.compute.ComputeV1.GcpResource
+
+DEFAULT_TEST_PORT: Final[int] = 8080
+DISCOVERY_URI: Final[str] = "https://run.googleapis.com/$discovery/rest?"
+
+
+@dataclasses.dataclass(frozen=True)
+class CloudRun:
```
sergiitk, line 33: This should be `CloudRunService`

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+    service_name: str
+    url: str
+
+    @classmethod
+    def from_response(cls, name: str, d: Dict[str, Any]) -> "CloudRun":
```
sergiitk, line 38:
```suggestion
    def from_response(cls, name: str, d: dict[str, Any]) -> "CloudRun":
```

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+
+from framework.infrastructure import gcp
+
+logger = logging.getLogger(__name__)
+
+# Type aliases
+GcpResource = gcp.compute.ComputeV1.GcpResource
+
+DEFAULT_TEST_PORT: Final[int] = 8080
+DISCOVERY_URI: Final[str] = "https://run.googleapis.com/$discovery/rest?"
```
sergiitk, line 29: this should be parsed from a flag so that we can override it for staging (or other) env

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+    def get_service_uri(self, service_name: str) -> str:
+        response = self.get_service(self.service, service_name)
+        return response.url
+
+    def delete_service(self, service_name: str):
+        # pylint: disable=no-member
+        self._delete_resource(
+            self.service.projects().locations().services(),
+            full_name=self.resource_full_name(
+                service_name, "services", self.region
```
sergiitk, line 110: make "services" a class constant, like in https://github.com/grpc/psm-interop/blob/12922820b1d662b8a1b7a11d9199c4c606a49923/framework/infrastructure/gcp/network_security.py#L160

sergiitk (changes requested):
please incorporate the fix for `KeyError: 'done'`

framework/infrastructure/gcp/api.py
```diff
@@ -574,7 +590,7 @@ def _execute(  # pylint: disable=arguments-differ
         timeout_sec: int = GcpProjectApiResource._WAIT_FOR_OPERATION_SEC,
     ):
         operation = request.execute(num_retries=self._GCP_API_RETRIES)
-        logger.debug("Operation %s", operation)
+        logger.info("Operation %s", operation)
```
sergiitk, line 593: please revert to debug

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,147 @@
...
+                        }
+                    ],
+                },
+            }
+
+            logger.info("Deploying Cloud Run service '%s'", service_name)
+            self.create_service(self.service, service_name, service_body)
+            return self.get_service_uri(service_name)
+
+        except Exception as e:  # noqa pylint: disable=broad-except
```
sergiitk, line 145: Here and elsewhere: avoid broad exceptions. I'm not sure we even need to catch anything here, just let the error propagate as is

kubernetes-manifests/client.deployment.yaml
sergiitk: To give some context here, at some point `client-deployment.yaml` and `client-secure.deployment.yaml` (same for the server) were much more different then now. Ideally we'd want to unify everything and use a single `client-secure.deployment.yaml` template.

Sometime ago I've figured out a good enough way to do it without pollution the template with irrelevant things too much, see https://github.com/grpc/psm-interop/blob/12922820b1d662b8a1b7a11d9199c4c606a49923/kubernetes-manifests/server.deployment.yaml#L1 and https://github.com/grpc/psm-interop/blob/12922820b1d662b8a1b7a11d9199c4c606a49923/kubernetes-manifests/prestop-hook.yaml.

I won't ask you to use combine these templates right now. I think, for now use whatever template is more convenient, and then we'll unify them.

framework/infrastructure/traffic_director_cloud_run.py
```diff
@@ -0,0 +1,87 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import logging
+from typing import Optional, Sequence
+
+from framework.infrastructure import gcp
+from framework.infrastructure.traffic_director import (
+    TrafficDirectorAppNetManager,
+)
```
sergiitk, line 20: Here and elsewhere: we're not importing anything from the file module (including classes), we alway import the module itself. See https://google.github.io/styleguide/pyguide.html#22-imports.

Note that there are a few exemptions, notably `typing`:

> Exemptions from this rule:
> Symbols from the following modules are used to support static analysis and type checking:
> - typing module
> - collections.abc module
> - typing_extensions module

sergiitk (changes requested):

framework/infrastructure/gcp/compute.py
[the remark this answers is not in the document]
sergiitk: Looks like this was fixed now.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,129 @@
...
+        """Returns the API name for Cloud Run."""
+        return "Cloud Run"
+
+    @property
+    def api_version(self) -> str:
+        """Returns the API version for Cloud Run."""
+        return "v2"
+
+    def create_service(self, service_name: str, body: dict) -> GcpResource:
+        return self._create_resource(
```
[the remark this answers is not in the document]
sergiitk: Fixed it in the other places: #168.

kubernetes-manifests/client.deployment.yaml
sergiitk: To make template merging a bit easier in the future, I've created the PR that cleans up GRPC_EXPERIMENTAL env vars: #169.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,129 @@
...
+    def from_response(
+        cls, name: str, response: dict[str, Any]
+    ) -> "CloudRunService":
+        return cls(
+            service_name=name,
+            url=response["urls"],
+        )
+
+
+class CloudRunApiManager(
```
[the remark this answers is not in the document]
sergiitk: Ah, yes, you're right.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,129 @@
...
+        if not project:
+            raise ValueError("Project ID cannot be empty or None.")
+        if not region:
+            raise ValueError("Region cannot be empty or None.")
+        # api_manager = gcp.api.GcpApiManager(v2_discovery_uri=DISCOVERY_URI)
+        self.region = region
+        super().__init__(api_manager.cloudrun(self.api_version), project)
+        self._services_collection = self.api.projects().locations().services()
+
+    SERVICES = "services"
```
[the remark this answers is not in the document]
sergiitk: Yes, this is totally fine. Just suggesting to move them up closer to the top of the class, similar to https://github.com/grpc/psm-interop/blob/8b8016ea3ae2bfb8cbade7d7521fbc9d2c2f0fee/framework/infrastructure/gcp/network_security.py#L136-L141

It's bad for readability when you have to search for constants in between methods :)

framework/infrastructure/gcp/cloud_run.py
sergiitk: The idea is that `TrafficDirectorAppNetManager` will eventually be renamed to `AppNetMeshManager` or `AppNetMeshManager`, as well as the rest of the classes. `TrafficDirectorManager` will be `TrafficDirectorMeshManager` too.

All these are resource managers holding the state of the mesh under test - among with any test-specific resources we need.

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +566,60 @@ def get_backend_service_backend_health(self, backend_service, backend):
             .execute()
         )
 
+    def create_serverless_neg(self, name: str, region: str, service_name: str):
```
sergiitk, line 569: The pattern here is `<verb>_<resource_type>_<resurce_subtype_if_any>`. For example, we have 
* create_url_map
* create_url_map_with_content.

We should rename `create_serverless_neg` to `create_neg_serverless`

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +566,60 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+        neg = self.get_serverless_network_endpoint_group(name, region)
+        logger.info("Created serverless NEG %s ", neg)
+        return neg
+
+    def delete_serverless_neg(self, name: str, region: str):
+        """Deletes a serverless NEG.
+
+        Args:
+            name: The name of the NEG to delete.
+            zone: The zone of the NEG.
```
sergiitk, line 605: region

framework/infrastructure/traffic_director_cloud_run.py
```diff
@@ -0,0 +1,119 @@
...
+        capacity_scaler: float = 1.0,
+        *,
+        circuit_breakers: Optional[dict[str, int]] = None,
+    ):
+        new_backends = []
+        for backend in backends:
+            new_backend = {
+                "group": backend,
+                "balancingMode": balancing_mode,
+                "maxRatePerEndpoint": max_rate_per_endpoint,
```
sergiitk, line 73: Since we're making a method specific to Cloud Run, we don't need to carry unnecessary fields. `maxRatePerEndpoint` was pretty much a required field to simply create proxyless backend service. Since you're making it None by default, I think we can drop it for good

framework/infrastructure/traffic_director_cloud_run.py
```diff
@@ -0,0 +1,119 @@
...
+                "capacityScaler": capacity_scaler,
+            }
+
+            if circuit_breakers is not None:
+                new_backend["circuitBreakers"] = circuit_breakers
```
sergiitk, line 78: do we know if this is even supported for serverless NEGs?

framework/infrastructure/gcp/compute.py
```diff
@@ -558,6 +566,60 @@ def get_backend_service_backend_health(self, backend_service, backend):
...
+        """
+        logger.info("Deleting serverless NEG %s in %s", name, region)
+        self._delete_resource(
+            self.api.regionNetworkEndpointGroups(),
+            resource_type="networkEndpointGroup",
+            resource_name=name,
+            region=region,
+        )
+
+    def get_serverless_network_endpoint_group(self, name, region):
```
sergiitk, line 615: let's make it `get_neg_serverless` for consistency. we probably should rename `get_network_endpoint_group` to `get_neg` too eventually (not in this PR tho)

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+Run xDS Test Server on Cloud Run.
+"""
+import logging
+from typing import List
+
+from typing_extensions import override
+
+from framework.infrastructure import gcp
+from framework.test_app.runners.cloud_run import cloud_run_base_runner
+from framework.test_app.server_app import XdsTestServer
```
sergiitk, line 24: you missed this one from my previous comment about imports

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Run xDS Test Server on Cloud Run.
+"""
+import logging
+from typing import List
```
sergiitk, line 18: todo - replace with `list[]` in code

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+            XdsTestServer(
+                ip="0.0.0.0", rpc_port=0, hostname=self.current_revision
+            )
+        ]
+        self.servers = servers  # Add servers to the list
+        self._start_completed()
+        return servers
+
+    def get_service_url(self):
+        return self.cloud_run_api_manager.get_service(self.service_name).url
```
sergiitk, line 79: maybe just return `CloudRunService`, and access its `.url`?

framework/test_cases/cloud_run_testcase.py
```diff
@@ -0,0 +1,167 @@
...
+import datetime as dt
+import logging
+from typing import Any, List, Optional
+
+from framework import xds_flags
+from framework import xds_k8s_testcase
+from framework.helpers import retryers
+from framework.infrastructure import k8s
+from framework.infrastructure import traffic_director
+from framework.infrastructure.gcp.compute import ComputeV1
```
sergiitk, line 23: same comment - imports

framework/xds_k8s_testcase.py
```diff
@@ -560,6 +562,20 @@ def _assertRpcsEventuallyGoToGivenServers(
...
+        want = frozenset(["endpoint_config"])
+        for generic_xds_config in config.generic_xds_configs:
+            if re.search(
+                r"\.ClusterLoadAssignment$", generic_xds_config.type_url
+            ):
+                seen.add("endpoint_config")
+        logger.debug(
+            "Received xDS config dump: %s",
+            json_format.MessageToJson(config, indent=2),
+        )
```
sergiitk, line 576: is this different than the log message `self.assertEDSConfigExists(config)` call below?

kubernetes-manifests/client.deployment.yaml
```diff
@@ -114,11 +125,14 @@ spec:
             - "--xds-server-uri=${xds_server_uri}"
             % endif
             % if config_mesh:
-            - "--config-mesh-experimental=${config_mesh}"
+            - "--config-mesh=${config_mesh}"
```
sergiitk, line 128: is this a required change? 
if no, I'd rather add it in a separate pr

kubernetes-manifests/client.deployment.yaml
```diff
@@ -54,8 +57,16 @@ spec:
           ports:
             - containerPort: ${stats_port}
           env:
+            - name: GRPC_TRACE
+              value: "xds_client"
```
sergiitk, line 61: this is cpp-specific, right?

framework/infrastructure/gcp/cloud_run.py
sergiitk: so where it is now?

framework/test_app/runners/k8s/k8s_xds_client_runner.py
```diff
@@ -131,12 +131,13 @@ def run(  # pylint: disable=arguments-differ
         log_to_stdout: bool = False,
         request_payload_size: int = 0,
         response_payload_size: int = 0,
+        is_trusted_xds_server_experimental: bool = False,
```
sergiitk, line 134: add `is_trusted_xds_server_experimental` to `ClientDeploymentArgs` instead

framework/test_app/runners/k8s/k8s_xds_client_runner.py
```diff
@@ -131,12 +131,13 @@ def run(  # pylint: disable=arguments-differ
...
     ) -> client_app.XdsTestClient:
         logger.info(
             (
                 'Deploying xDS test client "%s" to k8s namespace %s: '
                 "server_target=%s rpc=%s qps=%s metadata=%r secure_mode=%s "
-                "print_response=%s"
+                "print_response=%s is_trusted_xds_server_experimental=%s"
```
sergiitk, line 140: we only print actual arguments we pass to the binary here:

```
❯ docker run --interactive --tty --rm us-docker.pkg.dev/grpc-testing/psm-interop/java-client:v1.72.x --help
Usage: [ARGS...]

  --num_channels=INT     Default: 1
  --print_response=BOOL  Write RPC response to stdout. Default: false
  --qps=INT              Qps per channel, for each type of RPC. Default: 1
  --rpc=STR              Types of RPCs to make, ',' separated string. RPCs can be EmptyCall or UnaryCall. Default: UnaryCall
  --metadata=STR         The metadata to send with each RPC, in the format EmptyCall:key1:value1,UnaryCall:key2:value2.
  --rpc_timeout_sec=INT  Per RPC timeout seconds. Default: 20
  --server=host:port     Address of server. Default: localhost:8080
  --secure_mode=BOOLEAN  Use true to enable XdsCredentials. Default: false
  --request_payload_size=INT   Per-request size. Default: 0
  --response_payload_size=INT  Per-response size. Default: 0
  --enable_csm_observability=BOOL  Enable CSM observability reporting. Default: false
  --stats_port=INT       Port to expose peer distribution stats service. Default: 8081
```

framework/test_app/runners/k8s/k8s_xds_client_runner.py
```diff
@@ -191,6 +193,7 @@ def run(  # pylint: disable=arguments-differ
             config_mesh=config_mesh,
             generate_mesh_id=generate_mesh_id,
             print_response=print_response,
+            is_trusted_xds_server_experimental=is_trusted_xds_server_experimental,
```
sergiitk, line 196: This won't be needed one it's moved to `ClientDeploymentArgs`. The template will stay the same.

framework/test_cases/cloud_run_testcase.py
```diff
@@ -0,0 +1,167 @@
...
+                gcp_api_manager=self.gcp_api_manager,
+            )
+        test_servers = self.server_runner.run()
+        for test_server in test_servers:
+            test_server.set_xds_address(
+                self.server_xds_host, self.server_xds_port
+            )
+        return test_servers
+
+    def assertEDSConfigExists(self, config):
```
sergiitk, line 122: Add `@override` to make it clear we're replacing parent's assertion method. Unfortunately, we have too many of them to track.

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+            self.image_name,
+        )
+
+        super().run(**kwargs)
+        servers = [
+            XdsTestServer(
+                ip="0.0.0.0", rpc_port=0, hostname=self.current_revision
+            )
+        ]
+        self.servers = servers  # Add servers to the list
```
sergiitk, line 74: 1. `Instance attribute servers defined outside __init__ `
2. I don't think this is ever used

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+        self._start_completed()
+        return servers
+
+    def get_service_url(self):
+        return self.cloud_run_api_manager.get_service(self.service_name).url
+
+    @override
+    def cleanup(self, *, force=False):
+        try:
+            self.stop()
```
sergiitk, line 84: let's add a todo item to collect service logs to a file

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
```diff
@@ -0,0 +1,200 @@
...
+                service_name=self.service_name,
+                gcp_project=self.project,
+                gcp_ui_url=self.gcp_ui_url,
+                location=self.region,
+                start_time=run.time_start_requested,
+                cursor_time=run.time_start_completed,
+                end_time=run.time_stopped,
+            )
+
+    def stop(self):
```
sergiitk, line 197: rename this to `cleanup`, and accept `force: bool = False`  argument

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,86 @@
...
+            region=region,
+            gcp_api_manager=gcp_api_manager,
+        )
+        # Mutable state associated with each run.
+        self._reset_state()
+
+    @override
+    def _reset_state(self):
+        super()._reset_state()
+        self.service = None
```
sergiitk, line 55: you didn't save the service object - I think you should

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
```diff
@@ -0,0 +1,200 @@
...
+                f"Service {self.service_name}: start has already been "
+                f"requested at {self.time_start_requested.isoformat()}"
+            )
+
+        self._reset_state()
+        self.time_start_requested = _datetime.now()
+        self.current_revision = self.cloud_run_api_manager.deploy_service(
+            self.service_name,
+            self.image_name,
+        ).url
```
sergiitk, line 119: let's move this to the cloud run server runner, and then

1. store the full service object, not just url - as `CloudRunServerRunner.service`
2. clean it up as a part of `CloudRunServerRunner.cleanup`

sergiitk: note that `KubernetesBaseRunner.run` pretty much just deals with history tracking

eshitachandwani:
Include this in your PR:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/f4cdb38e-5a6b-4727-b10f-6f7f8f5a49ae)

eshitachandwani:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/b5ca9a07-fdc5-4293-8dca-5bdb07724eaa)
- [ ] [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/02eb14a4-226d-4cb9-b7ff-58b3137a107e)
- [ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/549900e6-0f47-4e82-a070-22f61b27bd74)

framework/infrastructure/gcp/compute.py
sergiitk: You missed `delete_serverless_neg` -> `delete_neg_serverless`

tests/cloud_run_csm_inbound_test.py
```diff
@@ -12,37 +12,56 @@
...
+        with self.subTest("3_create_backend_service"):
+            self.td.create_backend_service(
+                protocol=self.compute_v1.BackendServiceProtocol.HTTP2,
+            )
 
-        with self.subTest("2_finished"):
-            time.sleep(2)
+        with self.subTest("4_add_server_backends_to_backend_service"):
+            self.td.backend_service_add_cloudrun_backends(
+                [self.td.neg], self.region
```
sergiitk, line 45: the arguments are wrong (despite working somehow)

> I0416 17:00:21.987738 140704465681344 cloud_run_mesh_manager.py:79] Adding backends to Backend Service stk-cr-backend-service-20250416-2359-dx898: [{'group': ComputeV1.GcpResource(name='stk-cr-grpc-neg-20250416-2359-dx898', url='...'), 'balancingMode': 'us-central1', 'capacityScaler': 1.0}]

Note `'balancingMode': 'us-central1',`

You don't need to pass the region anymore. In fact, you don't need to pass the backends too.
`self.td` is a resource manager, and its purpose is to track resources your test created.

It already knows what about `td.neg`. Note that `backend_service_patch_backends` doesn't have the backends in method signature too

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Managed resources
+        self.grpc_route: Optional[GrpcRoute] = None
+        self.mesh: Optional[Mesh] = None
+        self.neg: Optional[GcpResource] = None
+
+    def backend_service_add_cloudrun_backends(
+        self,
+        backends: Sequence[str],
+        balancing_mode: str = "CONNECTION",
```
sergiitk, line 66: To prevent errors like passing the region for the value of `balancing_mode`, require optional arguments to be passed as kwargs:

```suggestion
        *,
        balancing_mode: str = "CONNECTION",
```

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,129 @@
...
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TEST_PORT: Final[int] = 8080
+
+
+class CloudRunServerRunner(cloud_run_base_runner.CloudRunBaseRunner):
+    """Manages xDS Test Servers running on Cloud Run."""
+
+    service: Optional[gcp.cloud_run.CloudRunV2] = None
```
sergiitk, line 34: you mean `CloudRunService`

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,129 @@
...
+        self._start_completed()
+        return servers
+
+    def deploy_service(
+        self,
+        service_name: str,
+        image_name: str,
+        *,
+        test_port: int = DEFAULT_TEST_PORT,
+    ):
```
sergiitk, line 96:
```suggestion
    ) -> gcp.cloud_run.CloudRunService:
```

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,129 @@
...
+                ],
+            },
+        }
+
+        logger.info("Deploying Cloud Run service '%s'", service_name)
+        self.cloud_run_api_manager.create_service(service_name, service_body)
+        return self.cloud_run_api_manager.get_service(service_name)
+
+    def get_service(self):
+        return self.cloud_run_api_manager.get_service(self.service_name)
```
sergiitk, line 119: remove - this is unused, and from the class perspective this should be returning `self.service`

tests/cloud_run_csm_inbound_test.py
```diff
@@ -12,37 +12,54 @@
...
-        with self.subTest("0_started"):
-            time.sleep(2)
+        with self.subTest("2_create_serverless_neg"):
+            self.td.create_neg_serverless(self.server_namespace)
 
-        with self.subTest("1_executing"):
-            time.sleep(2)
+        with self.subTest("3_create_backend_service"):
+            self.td.create_backend_service(
+                protocol=gcp.compute.ComputeV1.BackendServiceProtocol.HTTP2,
```
sergiitk, line 41: are we ever expecting this to support grpc backend? if so, let's add a todo with a bug or just a note

framework/infrastructure/gcp/compute.py
sergiitk: I think you missed this one

sergiitk (commented):
backend patching is broken

eshitachandwani:
Include this in your PR:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/feb79842-62fb-48b4-bf4c-af78167ad1e7)
- [ ] [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/c754e8a7-f60f-4b56-bfbd-ae2a51f2ad4a)
- [ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/0093b4ab-e6f5-478e-8c80-403100e8d244)
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/3c4a7419-c3dc-4d4c-96ed-ffa258294c7e)
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/98baa926-9fa3-4016-b88b-2d034326f490)

eshitachandwani:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/feb79842-62fb-48b4-bf4c-af78167ad1e7)
- [ ] [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/c754e8a7-f60f-4b56-bfbd-ae2a51f2ad4a)
- [ ] [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/0093b4ab-e6f5-478e-8c80-403100e8d244)
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/3c4a7419-c3dc-4d4c-96ed-ffa258294c7e)
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/98baa926-9fa3-4016-b88b-2d034326f490)

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,93 @@
...
+
+from framework.infrastructure import gcp
+
+logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass(frozen=True)
+class CloudRunService:
+    service_name: str
+    url: str
```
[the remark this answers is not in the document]
sergiitk: missed this one

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
```diff
@@ -0,0 +1,96 @@
...
+            self.service_name,
+            self.image_name,
+        )
+
+        super().run(**kwargs)
+        self.service = self.cloud_run_api_manager.deploy_service(
+            self.service_name,
+            self.image_name,
+        )
+        self.current_revision = self.service.url
```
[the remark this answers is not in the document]
sergiitk: missed this one

framework/infrastructure/gcp/cloud_run.py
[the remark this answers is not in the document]
sergiitk: nope, my branch was out of date.

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
```diff
@@ -0,0 +1,113 @@
...
+            new_backends,
+        )
+
+        self.compute.backend_service_patch_backends(
+            self.backend_service,
+            self.backends,
+            is_cloud_run=True,
+        )
+
+    def create_neg_serverless(self, service_name: str):
```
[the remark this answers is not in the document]
sergiitk: missed this one

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        balancing_mode: str = "CONNECTION",
+        capacity_scaler: float = 1.0,
+    ):
+        new_backends = []
+        for backend in self.backends:
+            new_backend = {
+                "group": backend,
+                "balancingMode": balancing_mode,
+                "capacityScaler": capacity_scaler,
+            }
```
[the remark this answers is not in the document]
sergiitk: take a look at this one once again

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+            new_backends.append(new_backend)
+
+        logging.info(
+            "Adding Cloud Run backends to Backend Service %s: %r",
+            self.backend_service.name,
+            new_backends,
+        )
```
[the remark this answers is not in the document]
sergiitk: See https://github.com/grpc/psm-interop/pull/154#discussion_r2049766969

framework/test_cases/cloud_run_testcase.py
```diff
@@ -0,0 +1,161 @@
...
+            network=self.network,
+            debug_use_port_forwarding=self.debug_use_port_forwarding,
+            enable_workload_identity=self.enable_workload_identity,
+            deployment_template="client.deployment.yaml",
+            stats_port=self.client_port,
+            reuse_namespace=self.server_namespace == self.client_namespace,
+            **kwargs,
+        )
+
+    def setupServerBackends(
```
sergiitk, line 83: could you please double check for me that this is used?

sergiitk (approved):
You missed a few things, but overall LGTM.

One of the things is looks like these fields are gone from `backend_service_add_cloudrun_backends` now - I don't think this was intentional
```
                "balancingMode": balancing_mode,
                "capacityScaler": capacity_scaler,
```

Also please remember to run the tests one more time before merging!

framework/test_cases/cloud_run_testcase.py
eshitachandwani: removed.

tests/cloud_run_csm_inbound_test.py
eshitachandwani: Need to cofirm with cloudrun team, will add in next PR if needed.

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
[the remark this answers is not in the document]
eshitachandwani: Right ! I missed the comment

framework/test_app/runners/cloud_run/cloud_run_xds_server_runner.py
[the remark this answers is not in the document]
eshitachandwani: No , it should be `uri` only  becuase `latestCreatedRevision` is just the name of the revision and we need the complete uri.

PR #160, opened 2025-03-11
skipping failover test due to switch to single region sandbox

sreenithi:
Context: We're shifting to creating a single region sandbox as recently we've been seeing repeated cluster repair operations on our multi regional sandbox, and the GKE team suggested using a single regional sandbox as a potential fix for the issue.  (Internal: b/386896489 and b/397859175)

Skipping failover test as it requires multi-region.

### DIFF
--- a/tests/failover_test.py
+++ b/tests/failover_test.py
@@ -33,6 +33,7 @@ _XdsTestServer = xds_k8s_testcase.XdsTestServer
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
 _KubernetesServerRunner = k8s_xds_server_runner.KubernetesServerRunner
 
+
 @absltest.skip("switching to a single regional sandbox due to b/397859175")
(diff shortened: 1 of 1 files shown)


tests/failover_test.py
```diff
@@ -34,6 +34,7 @@
 _KubernetesServerRunner = k8s_xds_server_runner.KubernetesServerRunner
 
 
+@absltest.skip("switching to a single regional sandbox due to b/397859175")
```
XuanWang-Amos, line 37: I think we should use existing framework function `is_supported` to skip this test:
https://github.com/grpc/psm-interop/blob/b753b6827d9cf9bef4fca0f68365f54c1af5a56c/tests/app_net_csm_observability_test.py#L176

XuanWang-Amos (commented):
Please add a test run to description.

sergiitk (changes requested):
If it's just affecting staging, we only need to skip it in staging.

PR #164, opened 2025-03-28
[CloudRun] Implement Cloudrun to Cloudrun test

eshitachandwani:
Tests the case where the client and the server are running in cloudrun. Verifies that 100 RPCs reach the server correctly and return with the correct response.


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -179,6 +179,7 @@ psm::cloud_run::setup() {
 psm::cloud_run::get_tests() {
   TESTS=(
     "cloud_run_csm_inbound_test"
+    "cloud_run_csm_outbound_test"
   )
```
kannanjgithub, line 183: It is not apparent from the test names in reference to what it is labelled as inbound and outbound. Can we call them gke_to_cloud_run_csm_test and cloud_run_to_cloud_run_csm_test?

bin/run_test_server_cloud_run.py
```diff
@@ -0,0 +1,51 @@
...
+
+def main(argv):
+    if len(argv) > 1:
+        raise app.UsageError("Too many command-line arguments.")
+
+    xds_flags.set_socket_default_timeout_from_flag()
+
+    run_kwargs = dict()
+    server_runner = common.make_cloud_run_server_runner()
+    server_runner.run(**run_kwargs)
```
XuanWang-Amos, line 47: If `run_kwargs` is just an `dict()`, can we just call `server_runner.run()` here?

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,232 @@
...
+DISCOVERY_URI: Final[str] = "https://run.googleapis.com/$discovery/rest?"
+
+
+@dataclasses.dataclass(frozen=True)
+class CloudRun:
+    service_name: str
+    url: str
+
+    @classmethod
+    def from_response(cls, name: str, d: Dict[str, Any]) -> "CloudRun":
```
XuanWang-Amos, line 39: Can we change `d` to a more meaningful name?

XuanWang-Amos (commented):
The kokoro run failed, please take a look.

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
```diff
@@ -0,0 +1,213 @@
...
+            gcp_ui_url=gcp_ui_url,
+            gcp_project=gcp_project,
+            query=query,
+            request=request,
+        )
+        link_to = service_name
+        # A whitespace at the end to indicate the end of the url.
+        logger.info("GCP Logs Explorer link to %s:\n%s ", link_to, link)
+
+    def logs_explorer_run_history_links(self):
```
kannanjgithub, line 184: It doesn't look like this is called from anywhere.

kannanjgithub (changes requested):
Review comments.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,232 @@
...
+        service.projects().locations().services().delete(
+            name=self.resource_full_name(service_name, "services", self.region)
+        ).execute()
+
+    def deploy_service(
+        self,
+        service_name: str,
+        image_name: str,
+        *,
+        test_port: int = DEFAULT_TEST_PORT,
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,232 @@
...
+        server_target: str = "",
+        mesh_name: str = "",
+    ):
+        if not service_name:
+            raise ValueError("service_name cannot be empty or None")
+        if not image_name:
+            raise ValueError("image_name cannot be empty or None")
+
+        try:
+            service_body = {
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,232 @@
...
+                                        "name": "GRPC_EXPERIMENTAL_XDS_AUTHORITY_REWRITE",
+                                        "value": "true",
+                                    },
+                                    {
+                                        "name": "GRPC_TRACE",
+                                        "value": "xds_client",
+                                    },
+                                    {
+                                        "name": "GRPC_VERBOSITY",
+                                        "value": "DEBUG",
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -0,0 +1,232 @@
...
+                                    },
+                                    {
+                                        "name": "is-trusted-xds-server-experimental",
+                                        "value": "true",
+                                    },
+                                    {
+                                        "name": "GRPC_XDS_BOOTSTRAP_CONFIG",
+                                        "value": "/tmp/grpc-xds/td-grpc-bootstrap.json",
+                                    },
+                                ],
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/rpc/grpc.py
```diff
@@ -95,10 +95,16 @@ def __init__(self, rpc_host):
         # Cache gRPC channels per port
         self.channels = dict()
 
-    def _make_channel(self, port) -> grpc.Channel:
+    def _make_channel(self, port, secure_mode=False) -> grpc.Channel:
         if port not in self.channels:
             target = f"{self.rpc_host}:{port}"
-            self.channels[port] = grpc.insecure_channel(target)
+            if secure_mode:
+                logger.info("eshita : secure mode detected")
```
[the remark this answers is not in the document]
eshitachandwani: done.

framework/test_app/runners/cloud_run/cloud_run_base_runner.py
eshitachandwani: It is called testcase cleanup to show link to cloud run logs in logs explorer
[here](https://github.com/grpc/psm-interop/pull/164/files#diff-0562d55945486687eb6462c4b7bc35be9eccc320294608618da8b3774a51b23dR1748-R1749)

kannanjgithub (approved):

XuanWang-Amos (approved):
Only comment left was the one for `run_kwargs`, otherwise LGTM.

In future PRs, please avoid including changes from other PRs, it's very hard to tell which part to review.

tests/cloud_run_csm_outbound_test.py
```diff
@@ -0,0 +1,77 @@
...
+_Lang: TypeAlias = skips.Lang
+_XdsTestServer: TypeAlias = xds_k8s_testcase.XdsTestServer
+_XdsTestClient: TypeAlias = xds_k8s_testcase.XdsTestClient
+
+
+class CloudRunCsmOutboundTest(cloud_run_testcase.CloudRunXdsTestCase):
+    @staticmethod
+    @override
+    def is_supported(config: skips.TestConfig) -> bool:
+        if config.client_lang is _Lang.CPP:
```
kannanjgithub, line 37: It should be 1.69.0 and add Python as well. Add Java as well with 1.71.0.

tests/gamma/affinity_test.py
```diff
@@ -51,6 +51,7 @@ def getClientRpcStats(
         num_rpcs: int,
         *,
         metadata_keys: Optional[tuple[str, ...]] = None,
+        secure_mode: Optional[bool] = False,
```
kannanjgithub, line 54: secuer_mode is not used in the method.

tests/gamma/affinity_session_drain_test.py
```diff
@@ -86,6 +86,7 @@ def getClientRpcStats(
         num_rpcs: int,
         *,
         metadata_keys: Optional[tuple[str, ...]] = None,
+        secure_mode: Optional[bool] = False,
```
kannanjgithub, line 89: Unused.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+
+from typing_extensions import override
+
+from framework.infrastructure import gcp
+from framework.test_app import client_app
+from framework.test_app.runners.cloud_run import cloud_run_base_runner
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CLIENT_TEST_PORT: Final[int] = 50052
```
kannanjgithub, line 28: Add a comment that this is the stats service port.

framework/test_app/client_app.py
```diff
@@ -109,6 +118,14 @@ def csds(self) -> _CsdsClient:
             log_target=f"{self.hostname}:{self.maintenance_port}",
         )
 
+    @property
+    @functools.lru_cache(None)
+    def secure_csds(self) -> _CsdsClient:
```
kannanjgithub, line 123: Add comment
When getting stats from the Cloud run client, there is a proxy involved which would use HTTP 1.1 for plaintext but  since we need HTTP 2.0 for gRPC to work, we need to only use encrypted communication.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+            "policy": {
+                "bindings": [
+                    {
+                        "role": "roles/run.invoker",
+                        "members": ["allUsers"],
+                    }
+                ],
+            },
+        }
+        self.cloud_run_api_manager.setIamPolicy(service_name, policy_body)
```
kannanjgithub, line 170: Make setIamPolicy failure throw exception instead of returning False so it is propagated to the testcase and fails the test.

tests/gamma/affinity_session_drain_test.py
eshitachandwani: Added to make the override method signature same as base class

tests/cloud_run_csm_outbound_test.py
eshitachandwani: Done.

framework/test_app/client_app.py
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
kannanjgithub: No, my comment was on the DEFAULT_CLIENT_TEST_PORT. That is the stats service port 50052.

tests/cloud_run_csm_outbound_test.py
```diff
@@ -0,0 +1,84 @@
...
+        with self.subTest("5_create_grpc_route"):
+            self.td.create_grpc_route(
+                self.server_xds_host, self.server_xds_port
+            )
+
+        with self.subTest("6_start_test_client"):
+            test_client: _XdsTestClient = self.startCloudRunTestClient(
+                test_server
+            )
+        # When getting stats from the Cloud Run client, there is a proxy
```
kannanjgithub, line 72: No, move the comment to "def secure_csds" method.

kannanjgithub (changes requested):
Follow-up comments.

framework/test_app/client_app.py
```diff
@@ -118,6 +123,11 @@ def csds(self) -> _CsdsClient:
             log_target=f"{self.hostname}:{self.maintenance_port}",
         )
 
+    # Secure mode is used when retrieving stats from the Cloud Run client. In
```
kannanjgithub, line 126: We can remove the last sentence and change the first sentence to: For fetching stats from gRPC client containers requiring secure communication, like Cloud run.

kannanjgithub (changes requested):
Nit.

kannanjgithub (approved):

tests/cloud_run_csm_inbound_test.py
```diff
@@ -38,6 +38,10 @@ class CloudRunCsmInboundTest(cloud_run_testcase.CloudRunXdsKubernetesTestCase):
     @override
     def is_supported(config: skips.TestConfig) -> bool:
         if config.client_lang is _Lang.CPP:
+            return config.version_gte("v1.69.x")
```
sergiitk, line 41: This is not directly related to the CR to CR tests, or framework changes made in this PR, right? If so, let's move it out to the separate PR and get it merged.

framework/xds_k8s_testcase.py
```diff
@@ -389,9 +389,15 @@ def removeServerBackends(self, *, server_runner=None):
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
     def assertSuccessfulRpcs(
-        self, test_client: XdsTestClient, num_rpcs: int = 100
+        self,
+        test_client: XdsTestClient,
+        num_rpcs: int = 100,
+        *,
+        secure_mode: Optional[bool] = False,
```
sergiitk, line 396: Is there a need for `undefined` state? If not, let's just make it bool False
```suggestion
        secure_mode:bool = False,
```

tests/gamma/affinity_test.py
sergiitk: let's add the `@override` annotation then

framework/xds_k8s_testcase.py
eshitachandwani: Done . In all places.

framework/test_cases/cloud_run_testcase.py
```diff
@@ -145,3 +149,52 @@ def tearDown(self):
...
+
+
+class CloudRunXdsTestCase(CloudRunXdsKubernetesTestCase):
+    client_runner: CloudRunClientRunner
+
+    @classmethod
+    def setUpClass(cls):
+        # Do not need to set up the Kubernetes API manager since both client and
+        # server run on Cloud Run.
+        cls.k8s_api_manager = None
```
sergiitk, line 161: We're extend `CloudRunXdsKubernetesTestCase` but we don't use Kubernetes?

framework/xds_k8s_testcase.py
```diff
@@ -239,9 +239,11 @@ def setUpClass(cls):
         cls.check_local_certs = _CHECK_LOCAL_CERTS.value
 
         # Resource managers
-        cls.k8s_api_manager = k8s.KubernetesApiManager(
-            xds_k8s_flags.KUBE_CONTEXT.value
-        )
+        if not hasattr(cls, "k8s_api_manager"):
```
sergiitk, line 242: This just seems wrong to me. We added it because `CloudRunXdsTestCase` doesn't use kubernetes. But it extends `CloudRunXdsKubernetesTestCase`, which extends this class, `XdsKubernetesBaseTestCase`. 

I mean it only makes sense to have `k8s_api_manager` in `XdsKubernetesBaseTestCase`, which implies kubernetes will be used.

sergiitk: Ideally we'd want to break up the classes so that there's a class that parses all the flags except kubernetes. Then have `XdsKubernetesBaseTestCase` that extends it and adds k8s-specific features. And have `CloudRunXdsTestCase` (name is up for discussion) that extends it - and doesn't use k8s at all.

However, I'll be ok if we made this change in a separate PR and only added todos in this one.
Keep in mind that the scripts that initiate test runs also assume we run on k8s - perform cluster auth, etc.

framework/test_cases/cloud_run_testcase.py
sergiitk: As I mentioned in the other comment, I'll accept a better solution with separating TestCase classes in a separate PR. However, relying on a a side-effect whether this class property is present is still a bad design.
Let's make it more explicit. Let's add a class method that determines whether to skip the k8s manager load.
If will return True by default in `XdsKubernetesBaseTestCase`, but this class will override it to return false.

Another option would be to keep the things as is - keep establishing cluster connection. As `psm_interop_kokoro_lib.sh` is right now, it will be initiating cluster authorization anyway - so the k8s context should be present and working.

eshitachandwani: Letting it remain as it is and adding Todos

eshitachandwani:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/d78aa055-7b2e-41c7-b249-565f2d2ef8e6)

framework/infrastructure/gcp/cloud_run.py
```diff
@@ -88,3 +88,12 @@ def delete_service(self, service_name: str):
                 service_name, self.SERVICES, self.region
             ),
         )
+
+    def setIamPolicy(self, service_name: str, policy_body: dict):
```
sergiitk, line 92:
```suggestion
    def set_iam_policy(self, service_name: str, policy_body: dict):
```

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+        super().__init__(
+            project,
+            service_name,
+            image_name,
+            network=network,
+            region=region,
+            gcp_api_manager=gcp_api_manager,
+        )
+
+        self._initalize_cloud_run_api_manager()
```
sergiitk, line 59: Here and in `CloudRunBaseRunner`: this shouldn't be called an API Manager `self.cloud_run_api_manager`. just `self.cloud_run_api`, or even `self.cloud_run`. 

In this framework Manager classes track resources under test. They are easy to identify - Managers always have cleanup methods to remove the resources at the end of the test run.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+
+        super().run(**kwargs)
+        self.service = self.deploy_service(
+            service_name=self.service_name,
+            image_name=self.image_name,
+            mesh_name=self.mesh_name,
+            server_target=self.server_target,
+        )
+        self.current_revision = self.service.uri
```
sergiitk, line 89: I still don't understand why do `self.current_revision = self.service.uri`. There's an actual field called revision. We use URI as a revision, but we never actually indicate it.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+        client_uri = self.current_revision.removeprefix("https://")
```
sergiitk, line 90: See the comment above. Wouldn't `client_uri = self.service.uri.removeprefix("https://")` make much more sense than `self.current_revision`? 

When I read `self.current_revision.removeprefix("https://")`, I think "why are we removing from "https://" from the revision? isn't the revision normally a sha digest or an autoincremented nubmer"?

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,181 @@
...
+        client = client_app.XdsTestClient(
+            ip=client_uri,
```
sergiitk, line 92: Will this be an actual IP? If not, better set to `0.0.0.0`.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,179 @@
...
+            "policy": {
+                "bindings": [
+                    {
+                        "role": "roles/run.invoker",
+                        "members": ["allUsers"],
+                    }
+                ],
+            },
+        }
+        self.cloud_run_api_manager.setIamPolicy(service_name, policy_body)
```
sergiitk, line 168: We have framework/infrastructure/gcp/iam.py, which provides handles Etag conflict resolution. Can you check if this makes sense?

framework/infrastructure/gcp/api.py
```diff
@@ -581,6 +581,29 @@ def _delete_resource(
...
 
+    def _set_iam_policy(
+        self,
+        collection: discovery.Resource,
+        body: dict,
+        full_name: str,
+        **kwargs,
+    ):
+        logger.info(
+            "Setting IAM policy for %s resource:\n%s",
```
sergiitk, line 592: we should log `full_name` here too

framework/rpc/grpc.py
```diff
@@ -95,10 +95,15 @@ def __init__(self, rpc_host):
         # Cache gRPC channels per port
         self.channels = dict()
 
-    def _make_channel(self, port) -> grpc.Channel:
+    def _make_channel(self, port, secure_mode=False) -> grpc.Channel:
```
sergiitk, line 98: I really appreciate this change you made! This looks great!

One thing I'm afraid of is that `secure_mode` might get conflated with security test's server's "secure_mode", where we listen on different ports - one with tls, another plain-text.

May I ask you to rename `secure_mode` to `secure_channel` here and elsewhere? Just the parameter, services like `secure_csds` are good.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,179 @@
...
+        return self.cloud_run_api_manager.get_service(service_name)
+
+    @override
+    def cleanup(self, *, force=False):
+        # TODO(emchandwani) : Collect service logs in a file.
+        try:
+            super().cleanup(force=force)
+            self.service = None
+            self.current_revision = None
```
sergiitk, line 177: You could call _reset_state here too. But I wonder if we already call it in super().cleanup or  self._stop - could you check please?

tests/cloud_run_csm_outbound_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+        with self.subTest("4_add_server_backends_to_backend_service"):
+            self.td.backend_service_add_cloudrun_backends()
+
+        with self.subTest("5_create_grpc_route"):
+            self.td.create_grpc_route(
+                self.server_xds_host, self.server_xds_port
+            )
+
+        with self.subTest("6_start_test_client"):
```
sergiitk, line 68:
```suggestion
        with self.subTest("6_start_cloudrun_test_client"):
```

similar to subtest 1

sergiitk (commented):
LGTM, but I'd like to have another look before merging

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -179,6 +179,7 @@ psm::cloud_run::setup() {
 psm::cloud_run::get_tests() {
   TESTS=(
     "cloud_run_csm_inbound_test"
+    "cloud_run_csm_outbound_test"
```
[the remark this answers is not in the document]
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Yes, added the tag.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Done.It was not called in either.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Should we remove the field? or add a comment?

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
eshitachandwani: Done.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,178 @@
...
+        logger.info("Deploying Cloud Run service '%s'", service_name)
+        self.cloud_run.create_service(service_name, service_body)
+        # Allow unauthenticated requests for `LoadBalancerStatsServiceClient`
+        # and `CsdsClient` to retrieve client statistics.
+        policy_body = {
+            "policy": {
+                "bindings": [
+                    {
+                        "role": "roles/run.invoker",
+                        "members": ["allUsers"],
```
sergiitk, line 163: Ok, I've looked in the details why it's needed and why the org policy for my project disallows it. 

This allows _public unauthenticated_ traffic to the service: https://cloud.google.com/run/docs/authenticating/public. If I understand this correctly, this pretty much publishes the service to the world. A big no-no. 

I've investigated further, and found a way to achieve the same without having to modify any IAM policies.
This page is what clued me in: https://cloud.google.com/run/docs/authenticating/developers#curl.

Using this approach, I was able to make grpcurl calls directly to the client service in my project without modifying IAMs.

```sh
❯ grpcurl -H "Authorization: Bearer $(gcloud auth print-identity-token)" stk-cr-client-20250501-1946-stldi-1040920224690.us-east7.run.app:443 grpc.channelz.v1.Channelz/GetTopChannels
{
  "channel": [
    {
      "ref": {
        "channel_id": "1"
...
```

Next, I was able to use it in the framework by providing call credentials when creating the channel:

```py
composite_credentials = grpc.composite_channel_credentials(
    grpc.ssl_channel_credentials(),
    grpc.access_token_call_credentials(IDENTITY_TOKEN),
)
self.channels[port] = grpc.secure_channel(
    target, credentials=composite_credentials
)
```

Where `IDENTITY_TOKEN` is hardcoded to the output of `gcloud auth print-identity-token`. With this, the framework was able to make direct calls to the CloudRun service. 

Obviously, we shouldn't be hardcoding the identity token. We should try to get it programmatically using google auth lib. I tried something like this, but it didn't work:

```py
request = google.auth.transport.requests.Request()
google_credentials: google.oauth2.credentials.Credentials
google_credentials, project_id = google.auth.default(
    request=request,
    # default_scopes=["https://www.googleapis.com/auth/cloud-platform"],
)
google_credentials.refresh(request)

logger.info(
    "Making call with credentials %s for project %s, expires %s",
    "'OK'" if google_credentials.token is not None else "'BAD'",
    project_id,
    google_credentials.expiry,
)

call_credentials = grpc.access_token_call_credentials(
    google_credentials.token
)
```

While this actually loaded the token (without the request and refresh it was None), it was different from `gcloud auth print-identity-token` and didn't work. I think there should be another way to do it. Maybe @gnossen or @XuanWang-Amos will be able to point you to the right direction.

Otherwise, I'm ok with loading it by getting the output of `gcloud auth print-identity-token` cli call.

I'll share the diff from my experimental branch directly with you.

sergiitk (changes requested):
We need to get rid of granting `roles/run.invoker` to `allUsers`. See my comment on `policy_body` with further explanations.

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
XuanWang-Amos: I don't have many context on auth but I think we can try this: `google.oauth2.id_token.fetch_id_token()`.

framework/rpc/grpc.py
```diff
@@ -95,10 +96,30 @@ def __init__(self, rpc_host):
...
-            self.channels[port] = grpc.insecure_channel(target)
+            if secure_channel:
+                result = subprocess.run(
+                    [
+                        "gcloud",
+                        "auth",
+                        "print-identity-token",
+                    ],
+                    check=True,
+                    capture_output=True,
```
sergiitk, line 110: If we go with this route, `timeout` is a must. IRRC, otherwise it'll be allowed to hang forever. Also, I'd like some logging before the command to indicate what's running.

framework/rpc/grpc.py
```diff
@@ -95,10 +96,30 @@ def __init__(self, rpc_host):
...
+                )
+                IDENTITY_TOKEN = result.stdout.decode("utf-8").strip()
```
sergiitk, line 112: Let's check that the token we've received is not empty.

framework/test_app/client_app.py
```diff
@@ -118,15 +146,23 @@ def get_load_balancer_stats(
...
     ) -> grpc_testing.LoadBalancerStatsResponse:
         """
         Shortcut to LoadBalancerStatsServiceClient.get_client_stats()
         """
-        return self.load_balancer_stats.get_client_stats(
-            num_rpcs=num_rpcs,
-            timeout_sec=timeout_sec,
-            metadata_keys=metadata_keys,
-        )
+        if secure_channel:
```
sergiitk, line 154: can be simplified to 

```py
lb_stats: LoadBalancerStatsServiceClient = self.secure_load_balancer_stats if secure_channel else self.load_balancer_stats
return lb_stats.get_client_stats( 
# ...
```

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
```diff
@@ -0,0 +1,165 @@
...
+            server_target=server_target,
+        )
+        self.current_revision = self.service.uri
+        client_uri = self.service.uri.removeprefix("https://")
+        client = client_app.XdsTestClient(
+            ip="0.0.0.0",
+            rpc_port=DEFAULT_PORT,
+            rpc_host=client_uri,
+            server_target=server_target,
+            hostname=self.current_revision,
```
sergiitk, line 96: `client_uri`?

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
sergiitk: What's the use of this field? 

If we save it, why not use `latestCreatedRevision` here? 
https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services#Service.FIELDS.latest_created_revision

In https://github.com/grpc/psm-interop/pull/154#discussion_r2053370576 you said
> No , it should be uri only becuase latestCreatedRevision is just the name of the revision and we need the complete uri.

But we're getting the uri from `self.service.uri` already.

framework/xds_k8s_testcase.py
```diff
@@ -572,8 +604,17 @@ def assertEDSConfigExists(self, config: ClientConfig):
                 seen.add("endpoint_config")
         self.assertSameElements(want, seen)
 
-    def assertXdsConfigExists(self, test_client: XdsTestClient):
-        config = test_client.csds.fetch_client_status(log_level=logging.INFO)
+    def assertXdsConfigExists(
+        self, test_client: XdsTestClient, *, secure_channel: bool = False
+    ):
+        if secure_channel:
```
sergiitk, line 610: same as in other place:
```py
csds: CsdsClient = test_client.secure_csds if secure_channel else test_client.csds
csds.fetch_client_status(
# ...
```

sergiitk (commented):
I'm getting the following error. 

```
Traceback (most recent call last):
  File "/Users/sergiitk/Development/psm-interop/tests/cloud_run_csm_outbound_test.py", line 74, in test_cloud_run_to_cloud_run
    self.assertXdsConfigExistsWithRetry(
  File "/Users/sergiitk/Development/psm-interop/framework/xds_k8s_testcase.py", line 591, in assertXdsConfigExistsWithRetry
    retryer(
  File "/Users/sergiitk/Development/psm-interop/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 423, in __call__
    do = self.iter(retry_state=retry_state)
  File "/Users/sergiitk/Development/psm-interop/venv/lib/python3.10/site-packages/tenacity/__init__.py", line 369, in iter
    return self.retry_error_callback(retry_state=retry_state)
  File "/Users/sergiitk/Development/psm-interop/framework/helpers/retryers.py", line 147, in error_handler
    raise RetryError(
framework.helpers.retryers.RetryError: Retry error calling framework.xds_k8s_testcase.XdsKubernetesBaseTestCase.assertXdsConfigExists: timeout 0:10:00 (h:mm:ss) exceeded. Last exception: AssertionError: unexpectedly None
Could not find correct bootstrap config before timeout 0:10:00 (h:mm:ss)
```

framework/test_app/runners/cloud_run/cloud_run_xds_client_runner.py
sergiitk: outdated

framework/test_cases/cloud_run_testcase.py
```diff
@@ -145,3 +149,53 @@ def tearDown(self):
...
+        test_client = self.client_runner.run(
+            server_target=test_server.xds_uri,
+            mesh_name=self.td.mesh.url,
+        )
+        return test_client
+
+    def cleanup(self):
+        self.server_runner.cleanup(force=self.force_cleanup)
+        self.td.cleanup(force=self.force_cleanup)
+        self.client_runner.cleanup(force=self.force_cleanup)
```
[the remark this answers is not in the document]
sergiitk: I see this is resolve - but I wonder if we should do the same thing in `CloudRunXdsKubernetesTestCase`

framework/rpc/grpc.py
```diff
@@ -95,10 +96,30 @@ def __init__(self, rpc_host):
...
+            if secure_channel:
+                result = subprocess.run(
+                    [
+                        "gcloud",
+                        "auth",
+                        "print-identity-token",
+                    ],
+                    check=True,
+                    capture_output=True,
+                )
```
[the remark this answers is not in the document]
sergiitk: See https://github.com/grpc/psm-interop/pull/164#discussion_r2074633031.

eshitachandwani:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/dec459d1-3b9c-4595-9b7f-a47241624685)

sergiitk:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/a5eb254a-abab-4f97-9479-77dfa39d275f)

sergiitk (approved):

eshitachandwani:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/b5d9a2db-a0c3-480d-90fb-01d348d03110)

- [ ] [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/5f870aa1-e07f-4ab1-b05d-2c4544435c14)

eshitachandwani:
- [ ] [grpc/core/master/linux/psm-cloud-run](https://source.cloud.google.com/results/invocations/43d3a39f-d30f-44c1-9e02-929b28925e6d)

PR #170, opened 2025-04-10
Dualstack test: fix the order of server namespace cleanup.

sergiitk:
Previously, `force_namespace=True` was passed to `self.v6_server_runner.cleanup()`, but the the call to this method is conditional on `if self.v6_server_runner:`. 

Therefore, when `self.v6_server_runner` isn't set, there won't be anything to call forced namespace cleanup. In practice, I don't see this happening as the code is right now, but still fixing the code for future-proofing (f.e. bad refactoring or copy-pasting to a different code flow).

### DIFF
--- a/tests/dualstack_test.py
+++ b/tests/dualstack_test.py
@@ -105,17 +105,17 @@ class DualStackTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
         self.client_runner.cleanup(
             force=self.force_cleanup, force_namespace=self.force_cleanup
         )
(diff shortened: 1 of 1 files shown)


sergiitk:
Test run: 

- [ ] [grpc/java/v1.72.x/branch/psm-dualstack](https://source.cloud.google.com/results/invocations/27d139ea-51d5-4487-9052-b378617f1d0a)

XuanWang-Amos (approved):

PR #171, opened 2025-04-16
[affinty_test] Add retry logic for asserting eds length

eshitachandwani:
The affinity_test is failing because server backends take time to ramp up to 3 or the number of replicas required , so when we request the status, we get only 1 endpoint. Adding a retry logic will give some time for backends to ramp up and only fail after a fixed timeout - 10 mins.

Bug : b/373265335

### DIFF
--- a/framework/xds_k8s_testcase.py
+++ b/framework/xds_k8s_testcase.py
@@ -388,40 +388,6 @@ class XdsKubernetesBaseTestCase(
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
-    def assertHealthyEndpointsCount(
-        self,
-        test_client,
-        expected_len,
-        msg=None,
-        *,
-        retry_timeout: dt.timedelta = dt.timedelta(minutes=3),
-        retry_wait: dt.timedelta = dt.timedelta(seconds=10),
-        log_level: int = logging.DEBUG,
-    ) -> None:
-        retryer = retryers.constant_retryer(
-            wait_fixed=retry_wait,
-            timeout=retry_timeout,
-            error_note=(
-                f"Timeout waiting for test client {test_client.hostname} to"
-                f" report {expected_len} endpoint(s)."
-            ),
-        )
-        for attempt in retryer:
-            with attempt:
-                client_config = test_client.csds.fetch_client_status_parsed(
-                    log_level=log_level
-                )
-                self.assertIsNotNone(
-                    client_config,
-                    "Error getting CSDS config dump"
-                    f" from client {test_client.hostname}",
-                )
-                self.assertLen(
-                    client_config.endpoints,
-                    expected_len,
-                    msg,
-                )
-
     def assertSuccessfulRpcs(
         self, test_client: XdsTestClient, num_rpcs: int = 100
     ) -> _LoadBalancerStatsResponse:
@@ -680,6 +646,41 @@ class XdsKubernetesBaseTestCase(
                 dumped_config,
             )
             raise retry_error
+
+    def assertHealthyEndpointsCount(
+        self,
+        test_client: XdsTestClient,
+        expected_count: int,
+        *,
+        retry_timeout: dt.timedelta = dt.timedelta(minutes=3),
+        retry_wait: dt.timedelta = dt.timedelta(seconds=10),
+        log_level: int = logging.DEBUG,
+    ) -> None:
+        retryer = retryers.constant_retryer(
+            wait_fixed=retry_wait,
+            timeout=retry_timeout,
+            error_note=(
+                f"Timeout waiting for test client {test_client.hostname} to"
+                f" report {expected_count} endpoint(s) in HEALTHY state."
+            ),
+        )
+        for attempt in retryer:
+            with attempt:
+                client_config = test_client.get_csds_parsed(log_level=log_level)
+                self.assertIsNotNone(
+                    client_config,
+                    "Error getting CSDS config dump"
+                    f" from client {test_client.hostname}",
+                )
+                logger.info(
+                    "<< Found EDS endpoints: HEALTHY: %s, DRAINING: %s",
+                    client_config.endpoints,
+                    client_config.draining_endpoints,
+                )
+                self.assertLen(
+                    client_config.endpoints,
+                    expected_count,
+                )
 
     def assertDrainingEndpointsCount(
         self,
--- a/framework/xds_k8s_testcase.py
+++ b/framework/xds_k8s_testcase.py
@@ -388,7 +388,9 @@ class XdsKubernetesBaseTestCase(
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
-    def assertEDSLenWithRetry(self, test_client, expected_len, msg=None) -> None:
+    def assertEDSLenWithRetry(
+        self, test_client, expected_len, msg=None
+    ) -> None:
         retryer = retryers.exponential_retryer_with_timeout(
             wait_min=dt.timedelta(seconds=10),
             wait_max=dt.timedelta(seconds=25),
--- a/framework/xds_k8s_testcase.py
+++ b/framework/xds_k8s_testcase.py
@@ -388,27 +388,39 @@ class XdsKubernetesBaseTestCase(
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
-    def assertEDSLenWithRetry(
-        self, test_client, expected_len, msg=None
+    def assertHealthyEndpointsCount(
+        self,
+        test_client,
+        expected_len,
+        msg=None,
+        *,
+        retry_timeout: dt.timedelta = dt.timedelta(minutes=3),
+        retry_wait: dt.timedelta = dt.timedelta(seconds=10),
+        log_level: int = logging.DEBUG,
     ) -> None:
-        retryer = retryers.exponential_retryer_with_timeout(
-            wait_min=dt.timedelta(seconds=10),
-            wait_max=dt.timedelta(seconds=25),
-            timeout=TD_CONFIG_MAX_WAIT,
-            log_level=logging.INFO,
+        retryer = retryers.constant_retryer(
+            wait_fixed=retry_wait,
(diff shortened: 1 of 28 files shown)


kannanjgithub (approved):

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,26 @@ def removeServerBackends(self, *, server_runner=None):
...
+            log_level=logging.INFO,
+            error_note=(
+                f"Did not get correct number of endpoints"
+                f" before timeout {TD_CONFIG_MAX_WAIT} (h:mm:ss)"
+            ),
+        )
+        retryer(self._assertEDSLen, test_client, expected_len, msg)
+
+    def _assertEDSLen(self, test_client, expected_len, msg=None):
+        parsed = test_client.csds.fetch_client_status_parsed(
```
purnesh42H, line 405: i think we just have to retry this and then do the assertion?

eshitachandwani: I dont understand what you mean by retry only this? As of now , we do get the status from client , but we get only 1 endpoint , so if we only keep fetching status with checking the assertion every time , how will we know when to stop? 
The change makes it so that we keep fetching the status and doing the assertion and retrying until the assertion passes or it times out.

purnesh42H: yeah but why do we need another intermediate method `assertEDSLen`? can't we just wrap self.assertLen in retryer in the file affinity_test.py? we just need to verify no. of servers and status being None is already verified?

purnesh42H: oh you need to fetch csds status as well. So, you just those 2 is enough but we don't another intermediate method

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,26 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertEDSLen(self, test_client, expected_len, msg=None) -> None:
+        retryer = retryers.exponential_retryer_with_timeout(
```
purnesh42H, line 392: Should we put a docstring here explaining why it needs to do retries. Similar to this https://github.com/grpc/psm-interop/blob/main/framework/test_cases/session_affinity_mixin.py#L71

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,26 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertEDSLen(self, test_client, expected_len, msg=None) -> None:
```
XuanWang-Amos, line 391: This function basically just use retryer to call `_assertEDSLen`, maybe it's better to use a name like `assertEDSLenWithRetry`

XuanWang-Amos:
> Could you add the successful kokoro run for the test?

+1

framework/xds_k8s_testcase.py
eshitachandwani: Done.

framework/xds_k8s_testcase.py
sergiitk: I'm all for docstrings, but that method in session affinity mixin only needed such extensive explanation because it's test-specific and describes infrastructure nuances. 
For something as trivial as asserting the length of EDS endpoints, I'd not require the docstring (though it's still encourageable).

We could put something like "Asserts the length of EDS endpoints retrieved via CSDS".

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,26 @@ def removeServerBackends(self, *, server_runner=None):
...
+            timeout=TD_CONFIG_MAX_WAIT,
+            log_level=logging.INFO,
+            error_note=(
+                f"Did not get correct number of endpoints"
+                f" before timeout {TD_CONFIG_MAX_WAIT} (h:mm:ss)"
+            ),
+        )
+        retryer(self._assertEDSLen, test_client, expected_len, msg)
+
+    def _assertEDSLen(self, test_client, expected_len, msg=None):
```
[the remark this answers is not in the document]
sergiitk: Not sure what you mean here

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,28 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertEDSLenWithRetry(
```
sergiitk, line 391: [Sidenote] Abbreviations should be converted to the camel case as well: `assertEdsLenWithRetry`

sergiitk (changes requested):
I recommend just copy-pasting `assertDrainingEndpointsCount`, naming it `assertHealthyEndpointsCount`, and just modifying it to only consider `client_config.endpoints` (remove all mentions of client_config.draining_endpoints )

framework/xds_k8s_testcase.py
eshitachandwani: Did not add the docstring because it feels redundant since name feels clear enough to me. Let me know if I should add it.

framework/xds_k8s_testcase.py
eshitachandwani: Noted.

eshitachandwani:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/c38a847a-ead4-4bc0-adf0-598811370fb4)

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,40 @@ def removeServerBackends(self, *, server_runner=None):
...
+            error_note=(
+                f"Timeout waiting for test client {test_client.hostname} to"
+                f" report {expected_len} endpoint(s)."
+            ),
+        )
+        for attempt in retryer:
+            with attempt:
+                client_config = test_client.csds.fetch_client_status_parsed(
+                    log_level=log_level
+                )
```
sergiitk, line 413: There's a shortcut that abstracts out the source of the data:

```suggestion
                client_config = test_client.get_csds_parsed(log_level=log_level)
```

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,40 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertHealthyEndpointsCount(
+        self,
+        test_client,
+        expected_len,
+        msg=None,
```
sergiitk, line 395: Is there a need for msg? see assertDrainingEndpointsCount. It's better to provide the default message, same as here: https://github.com/grpc/psm-interop/blob/8b8016ea3ae2bfb8cbade7d7521fbc9d2c2f0fee/framework/xds_k8s_testcase.py#L677-L681

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,40 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertHealthyEndpointsCount(
+        self,
+        test_client,
+        expected_len,
```
sergiitk, line 394: Please add the type hints, like here: https://github.com/grpc/psm-interop/blob/8b8016ea3ae2bfb8cbade7d7521fbc9d2c2f0fee/framework/xds_k8s_testcase.py#L649-L650

sergiitk, line 394: rename expected_len to expected_count to be consistent with the method name

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,40 @@ def removeServerBackends(self, *, server_runner=None):
...
+        for attempt in retryer:
+            with attempt:
+                client_config = test_client.csds.fetch_client_status_parsed(
+                    log_level=log_level
+                )
+                self.assertIsNotNone(
+                    client_config,
+                    "Error getting CSDS config dump"
+                    f" from client {test_client.hostname}",
+                )
```
sergiitk, line 418: it'll be useful to track the health over time with the log message like this: https://github.com/grpc/psm-interop/blob/8b8016ea3ae2bfb8cbade7d7521fbc9d2c2f0fee/framework/xds_k8s_testcase.py#L672-L676

framework/xds_k8s_testcase.py
```diff
@@ -386,6 +388,40 @@ def removeServerBackends(self, *, server_runner=None):
         # Remove backends from the Backend Service
         self.td.backend_service_remove_neg_backends(neg_name, neg_zones)
 
+    def assertHealthyEndpointsCount(
```
sergiitk, line 391: Please place the method just before `assertDrainingEndpointsCount` since they're so similar

eshitachandwani:
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/024ff563-5520-45fc-ac5a-5c5de193e929)

framework/xds_k8s_testcase.py
```diff
@@ -653,6 +653,41 @@ def assertRouteConfigUpdateTrafficHandoff(
...
+        )
+        for attempt in retryer:
+            with attempt:
+                client_config = test_client.get_csds_parsed(log_level=log_level)
+                self.assertIsNotNone(
+                    client_config,
+                    "Error getting CSDS config dump"
+                    f" from client {test_client.hostname}",
+                )
+                logger.info(
```
sergiitk, line 681: adding a note for future

```suggestion
                # TODO(nice to have): parse and log all health statuses.
                # https://github.com/envoyproxy/envoy/blob/b6df9719/api/envoy/config/core/v3/health_check.proto#L35
                logger.info(
```

sergiitk (approved):
looks good, thank you

sergiitk:
- [ ] [grpc/go/v1.72.x/branch/xds_k8s_lb](https://source.cloud.google.com/results/invocations/33d78c4d-aba9-4329-b127-aa12c4d4487d)

PR #173, opened 2025-04-24
cleanup.py: cleanup appnet traffic director resources

eshitachandwani:


sergiitk:
I'm not sure if this will work. We don't differentiate between AppNet and regular resources with prefixes. On the other hand, we don't seem to do different for PSM Security-specific resources too... Let's give it a run.

eshitachandwani:
- [ ] [grpc/core/master/linux/grpc_xds_resource_cleanup](https://source.cloud.google.com/results/invocations/34af0847-d90f-4fa3-b66c-910a40fb140a)

eshitachandwani:
> I'm not sure if this will work. We don't differentiate between AppNet and regular resources with prefixes. On the other hand, we don't seem to do different for PSM Security-specific resources too... Let's give it a run.

Attatched a cleanup run job , looks like its working. Its becuase the error was backend services were still using grpc route and grpc route is an appnet td resource.

kannanjgithub:
(Not directly related to this PR but related to the broader discussion on cleanup job failing) In the failing cleanup [job](https://fusion2.corp.google.com/invocations/e4415972-0a06-4694-936e-162367967056/targets/grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_xds_resource_cleanup/log) it complains about unable to delete namespace psm-interop-server-20250316-1130-nzdoa in the cluster cleanup for lb primary but I actually dont' find this namespace on that cluster. The only namespaces with psm-interop-server prefix are with 20250425 date.

kannanjgithub (approved):

eshitachandwani:
> (Not directly related to this PR but related to the broader discussion on cleanup job failing) In the failing cleanup [job](https://fusion2.corp.google.com/invocations/e4415972-0a06-4694-936e-162367967056/targets/grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_xds_resource_cleanup/log) it complains about unable to delete namespace psm-interop-server-20250316-1130-nzdoa in the cluster cleanup for lb primary but I actually dont' find this namespace on that cluster. The only namespaces with psm-interop-server prefix are with 20250425 date.

I think thats why it is failing also , in the log it says the `get` function is failing , so even the script is not able to find it !

bin/cleanup/cleanup.py
```diff
@@ -373,21 +373,21 @@ def cleanup_td_for_gke(
...
+    )
 
     logger.info(
         "----- Removing traffic director for gke, prefix %s, suffix %s",
         prefix,
         suffix,
     )
     security_td.cleanup(force=True)
-    # appnet_td.cleanup(force=True)
+    appnet_td.cleanup(force=True)
```
sergiitk, line 390: let's add comments like this before all `.cleanup()` calls here so it's easier for us to differentiate between resource manager classes if something goes wrong.
```py
logger.info("--- Cleaning up %s resources...", appnet_td.__class__.__name__)
```

sergiitk (approved):

bin/cleanup/cleanup.py
eshitachandwani: Done.

PR #177, opened 2025-05-15
[DRAFT] Round robin Test fix

pawbhard:


framework/xds_k8s_testcase.py
```diff
@@ -411,6 +411,55 @@ def assertSuccessfulRpcs(
         )
         return lb_stats
 
+    # Return number of servers using subchannels
+    def getActiveNumberOfServers(
```
pawbhard, line 415: @sergiitk , This is what we were discussing. Objective is to somehow find number of active servers before starting the tests which depend on this number

framework/xds_k8s_testcase.py
```diff
@@ -411,6 +411,55 @@ def assertSuccessfulRpcs(
         )
         return lb_stats
 
+    # Return number of servers using subchannels
```
sergiitk, line 414: I don't think we can derive it from the number of subchannels. One of the reasons: b/284497334. I'll think if this will make sense in a general case - after this issue is fixed.

framework/xds_k8s_testcase.py
sergiitk: Could you please clarify what you mean by "active"?

framework/xds_k8s_testcase.py
```diff
@@ -411,6 +411,55 @@ def assertSuccessfulRpcs(
...
+                f"Didn't find the expected number of active servers {expected_num_servers}"
+                f" before timeout {retry_timeout} (h:mm:ss)"
+            ),
+        )
+        retryer(
+            self._assertActiveNumberOfServers,
+            expected_num_servers,
+            test_client,
+            secure_channel,
+        )
```
sergiitk, line 460: FYI - In cases like this it's more convenient to use retryer's context manager feature instead of defining a separate method just for a single assert. Example: https://github.com/grpc/psm-interop/blob/b57b402246cee51058cdb4c41dd80d72ab0050e5/framework/xds_k8s_testcase.py#L711-L718

framework/xds_k8s_testcase.py
pawbhard: By active I meant healthy, we already had function so using that now. resolving this comment

kannanjgithub (approved):

tests/round_robin_test.py
```diff
@@ -63,7 +63,10 @@ def test_round_robin(self) -> None:
         with self.subTest("09_test_server_received_rpcs_from_test_client"):
             self.assertSuccessfulRpcs(test_client)
 
-        with self.subTest("10_round_robin"):
+        with self.subTest("10_test_healthy_endpoint_count"):
```
eshitachandwani, line 66: Not blocking but just a thought , it feels like the healthy endpoints should be verified even before asserting successful RPCs i.e 
before 
```
with self.subTest("09_test_server_received_rpcs_from_test_client"):
            self.assertSuccessfulRpcs(test_client)
```

pawbhard: Ack, We don't need step 9 after this. Would like to see if this fixes the flakiness before cleaning up 9

sergiitk:
- [ ] [grpc/core/master/linux/psm-dualstack](https://source.cloud.google.com/results/invocations/25d32dcf-8413-4504-a1bf-ce7a31b13e20)

sergiitk (approved):
LGTM, but please add internal metabug ref b/ to the PR description.

PR #178, opened 2025-05-19
Fix flaky of Affinity session drain test

pawbhard:
Removed the step ""04_confirm_all_servers_receive_traffic" . 

Issue: sometimes all servers don't receive traffic as some are created in another zone with priority 1

This test will be able to check session affinity without this step as we assert on number of RPC after setting Cookie.
Also increasing replica count so we have more chances of getting greater than 1 server in a subzone.

Only following case will be able to skip test.
It is possible that only 1 server is created in priority 0 . In this case our test would still pass but it will not confirm the feature is working correctly.

### DIFF
--- a/tests/gamma/affinity_session_drain_test.py
+++ b/tests/gamma/affinity_session_drain_test.py
@@ -118,23 +118,7 @@ class AffinitySessionDrainTest(  # pylint: disable=too-many-ancestors
         with self.subTest("03_start_test_client"):
             test_client = self.startTestClient(test_servers[0])
 
-        with self.subTest("04_confirm_all_servers_receive_traffic"):
-            # Note: the output of this call may or may not print out the
-            # cookie. This is *not* any kind of signal, just depends on system
-            # latency. This subtest does not use any cookies.
-            self.assertRpcsEventuallyGoToGivenServers(
-                test_client,
-                test_servers,
-                num_rpcs=120,  # Nice and even.
-                retry_timeout=TRAFFIC_PIN_TIMEOUT,
-                retry_wait=TRAFFIC_PIN_RETRY_WAIT,
-            )
-            logger.info(
-                "Confirmed all servers received traffic: %s",
-                [server.hostname for server in test_servers],
-            )
-
-        with self.subTest("05_retrieve_cookie"):
+        with self.subTest("04_retrieve_cookie"):
             result = self.assertSsaCookieAssigned(test_client, test_servers)
             cookie: Final[str] = result[0]
             chosen_server: Final[server_app.XdsTestServer] = result[1]
@@ -142,7 +126,7 @@ class AffinitySessionDrainTest(  # pylint: disable=too-many-ancestors
             chosen_name: Final[str] = chosen_server.hostname
             logger.info("Chosen server: %s, cookie: %s", chosen_name, cookie)
 
-        with self.subTest("06_only_chosen_server_receives_rpcs_with_cookie"):
+        with self.subTest("05_only_chosen_server_receives_rpcs_with_cookie"):
             logger.info("Configuring client to send cookie %s", cookie)
             test_client.update_config.configure_unary(
                 metadata=(
@@ -157,7 +141,7 @@ class AffinitySessionDrainTest(  # pylint: disable=too-many-ancestors
             )
             logger.info("Confirmed all RPCs went to %s", chosen_name)
 
-        with self.subTest("07_stopping_chosen_server"):
+        with self.subTest("06_stopping_chosen_server"):
             self.server_runner.delete_pod(
                 chosen_name,
                 grace_period=TERMINATION_GRACE_PERIOD,
@@ -165,10 +149,10 @@ class AffinitySessionDrainTest(  # pylint: disable=too-many-ancestors
                 wait_for_deletion=False,
             )
 
-        with self.subTest("08_test_client_csds_shows_chosen_server_draining"):
+        with self.subTest("07_test_client_csds_shows_chosen_server_draining"):
             self.assertDrainingEndpointsCount(test_client, 1)
 
-        with self.subTest("09_pinned_traffic_stays_on_draining_server"):
+        with self.subTest("08_pinned_traffic_stays_on_draining_server"):
             self.assertRpcsEventuallyGoToGivenServers(
                 test_client,
                 (chosen_server,),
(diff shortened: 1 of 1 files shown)


tests/gamma/affinity_session_drain_test.py
```diff
@@ -36,7 +36,7 @@
 _Lang: TypeAlias = skips.Lang
 
 # Constants.
-REPLICA_COUNT: Final[int] = 3
+REPLICA_COUNT: Final[int] = 6
```
kannanjgithub, line 39: Let's go with 4 for the replica count to reduce resource usage a bit.

kannanjgithub (approved):

eshitachandwani (approved):

sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/2eb36c2e-5de9-44f3-a3f6-4272a6570fab)

pawbhard:
> I thought about it a bit more, and I think we shouldn't change `TRAFFIC_PIN_TIMEOUT` because it's used in other checks too. Instead, in step 04 let's call `assertRpcsEventuallyGoToGivenServers` without `retry_timeout` argument, which will default it to 10 minutes. We should keep `retry_wait` however, by default it's just 1 second.

Will that solve the issue as priority 1 pods will not converge to pri 0, hence will not receive traffic. I last time increased the TIMEOUT which didn't help .
https://github.com/grpc/psm-interop/pull/158/files

pawbhard:
Test run : 
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/bea416b5-87b5-4b4d-8989-533969a4f070)

sergiitk (changes requested):
Discussed in chat: we do need step 04 to establish the baseline. Without it, the test won't be reliable because we don't establish that there's more than one servers receiving traffic in the first place. These two case will look identical to us:

1. One server getting traffic and the affinity behavior is broken.
2. One server getting traffic and the affinity behavior works as expected.

However, to make the test more resilient, we don't need to check for _all_ servers to receive traffic in subtest `04`. Confirming that it's more than one server is enough.

pawbhard:
> Discussed in chat: we do need step 04 to establish the baseline. Without it, the test won't be reliable because we don't establish that there's more than one servers receiving traffic in the first place. These two case will look identical to us:
> 
> 1. One server getting traffic and the affinity behavior is broken.
> 2. One server getting traffic and the affinity behavior works as expected.
> 
> However, to make the test more resilient, we don't need to check for _all_ servers to receive traffic in subtest `04`. Confirming that it's more than one server is enough.

Requested change has been done, please rer-eview
Passed run with latest changes 
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/8b20c119-a736-41ad-90b1-6fd888f994ed)

pawbhard:
Final RUn:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/b9e25e53-af49-45fc-8bf5-255634a65528)

framework/xds_k8s_testcase.py
```diff
@@ -517,6 +517,52 @@ def assertRpcStatusCodes(
...
+                f" {num_expected_servers} server(s)"
+                f" before timeout {retry_timeout} (h:mm:ss)"
+            ),
+        )
+        retryer(
+            self._assertRpcsEventuallyReachMinServers,
+            test_client,
+            num_expected_servers,
+            num_rpcs,
+        )
```
sergiitk, line 544: No need for an extra methods just for this. Use `with attempt` approach

```suggestion
        for attempt in retryer:
            with attempt:
                lb_stats = self.getClientRpcStats(test_client, num_rpcs)
                #... the rest of _assertRpcsEventuallyReachMinServers
```

This also improves the error message - instead of something like

`framework.helpers.retryers.RetryError: Retry error calling __main__.FakeTest._assertRpcsEventuallyReachMinServers: timeout 0:00:02 (h:mm:ss) exceeded. ... ` 
you'll get 
`framework.helpers.retryers.RetryError: Retry error: timeout 0:00:02 (h:mm:ss) exceeded. ... `

framework/xds_k8s_testcase.py
```diff
@@ -517,6 +517,52 @@ def assertRpcStatusCodes(
...
+            failed,
+            0,
+            msg=f"Expected all RPCs to succeed: {failed} of {num_rpcs} failed",
+        )
+        self.assertLessEqual(
+            num_expected_servers,
+            len(lb_stats.rpcs_by_peer),
+            msg=f"Rpc went to {lb_stats.rpcs_by_peer} servers expected"
+            f" at least {num_expected_servers} servers",
+        )
```
sergiitk, line 564: 1. The order of params in python assert is usually `actual, expected, msg`
2. Did you mean `len(lb_stats.rpcs_by_peer)`? Or you wanted to print hostnames? If so, we should still print the length. 

```suggestion
        self.assertGreaterEqual(
            len(lb_stats.rpcs_by_peer),
            num_expected_servers,
            msg=(
                f"RPCs went to {len(lb_stats.rpcs_by_peer)} server(s)"
                f" expected at least {num_expected_servers} servers:"
                f" {set(lb_stats.rpcs_by_peer.keys())}"
            ),
        )
```

Resulting error will look like this

```
framework.helpers.retryers.RetryError: Retry error: timeout 0:00:02 (h:mm:ss) exceeded. Last exception: AssertionError: 2 not greater than or equal to 3 : RPCs went to 2 server(s) expected at least 3 servers:  {'server-1', 'server-2'}
RPCs (num_rpcs=100) did not go to at least 3 server(s) before timeout 0:00:02 (h:mm:ss)
# [FakeTest.test_me] PSM Failed Test Traceback END
```

framework/xds_k8s_testcase.py
```diff
@@ -517,6 +517,52 @@ def assertRpcStatusCodes(
...
         )
 
+    def assertRpcsEventuallyReachMinServers(
+        self,
+        test_client: XdsTestClient,
+        num_expected_servers: int,
+        num_rpcs: int = 100,
+        *,
+        retry_timeout: dt.timedelta = TD_CONFIG_MAX_WAIT,
+        retry_wait: dt.timedelta = dt.timedelta(seconds=1),
```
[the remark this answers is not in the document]
pawbhard: changed

framework/xds_k8s_testcase.py
pawbhard: Done

framework/xds_k8s_testcase.py
pawbhard: Done, wanted to print length

pawbhard:
Final Run:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/06e09c72-3707-424d-bcf4-fb6e7e668d0e)

sergiitk (approved):

PR #179, opened 2025-05-22
Rename fallback test suite to light

murgatroid99:
The light test suite is for tests that don't use TD or GKE, and run the entire test on the same machine. Currently, the only member of the suite is the fallback test, but a federation test will be added soon.


sergiitk (approved):
Note that we'll need to make that build configs (in the repo) refer to the new name even in the old branches - because they all pull the most recent version of the install lib. So we'll have to backport.

sergiitk:
Internal ref cl/762145914

PR #181, opened 2025-06-02
Reorder checks for RPC status codes

pawbhard:
Earlier we were checking unwanted status and then the correct expected code.
Reordering this to check if we get expected status and then check for any unwanted status.

Reason: We are seeing a Flake where we get unwanted status b/416171553 . We wanted to create the matcher to ignore unwanted status, hence it make sense to swap checks as mentioned in cl/758143043

### DIFF
--- a/framework/xds_k8s_testcase.py
+++ b/framework/xds_k8s_testcase.py
@@ -491,7 +491,19 @@ class XdsKubernetesBaseTestCase(
             diff_stats, ignore_empty=True, highlight=False
         )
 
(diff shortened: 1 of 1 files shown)


kannanjgithub (approved):

pawbhard:
- [ ] [grpc/java/master/branch/psm-security](https://source.cloud.google.com/results/invocations/fc934fbd-3edc-4415-95a9-2f583e0d6f52)

sergiitk (approved):

PR #188, opened 2025-07-07
Add retries with waits in fallback mid_update test

kannanjgithub:
Adding retry with wait checks when looking for rpcs to go to the server indicated by the fallback control plane, and when expecting the ADS stream to the primary get reinstated.

Local success run: https://drive.google.com/file/d/108Wknp7_UZS_Lj7hCdfsW0GzcsR-C8qG/view?usp=sharing&resourcekey=0-Tsd7890oa_QMjXxAmG8szw

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 Mako~=1.1
 PyYAML~=6.0
 absl-py~=2.1
-black
 docker==7.1.0
 google-api-python-client~=1.12
 google-cloud-secret-manager~=2.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,6 @@ six~=1.13
 tenacity~=6.2
 typing-extensions~=4.10
 packaging~=23.1
-pylint
 Pygments~=2.9
 python-dateutil~=2.8
 protobuf~=4.24
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ six~=1.13
 tenacity~=6.2
 typing-extensions~=4.10
 packaging~=23.1
+pylint
 Pygments~=2.9
 python-dateutil~=2.8
 protobuf~=4.24
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 Mako~=1.1
 PyYAML~=6.0
 absl-py~=2.1
+black
 docker==7.1.0
 google-api-python-client~=1.12
 google-cloud-secret-manager~=2.1
--- a/tests/fallback_test.py
+++ b/tests/fallback_test.py
@@ -14,6 +14,7 @@
 import datetime
 import logging
 import socket
+import time
 
 import absl
 from absl import flags
@@ -311,6 +312,7 @@ class FallbackTest(absltest.TestCase):
                 primary_status=channelz_pb2.ChannelConnectivityState.TRANSIENT_FAILURE,
                 fallback_status=channelz_pb2.ChannelConnectivityState.READY,
             )
+            time.sleep(20)
             stats = client.get_stats(10)
             self.assertEqual(stats.num_failures, 0)
             self.assertIn("server2", stats.rpcs_by_peer)
@@ -320,6 +322,7 @@ class FallbackTest(absltest.TestCase):
                 port=self.bootstrap.primary_port,
                 upstream_port=server3.port,
             ):
+                time.sleep(60)
                 self.assert_ads_connections(
                     client,
                     primary_status=channelz_pb2.ChannelConnectivityState.READY,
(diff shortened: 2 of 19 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 Mako~=1.1
 PyYAML~=6.0
 absl-py~=2.1
+black
```
sergiitk, line 4: We don't need to install black when we _run_ the framework. Dev dependencies captured in requirements-dev.txt.

kannanjgithub: Removed.

requirements.txt
```diff
@@ -14,6 +14,7 @@ six~=1.13
 tenacity~=6.2
 typing-extensions~=4.10
 packaging~=23.1
+pylint
```
sergiitk, line 17: Please remove this too. Pylint in a bad shape now so it's not even added to dev dependencies. To run locally, use https://github.com/grpc/psm-interop/blob/main/bin/pylint.sh for now. I'll get it updated some time soon.

tests/fallback_test.py
```diff
@@ -133,7 +134,19 @@ def assert_ads_connections(
...
+            self.ads_connections_status_check_result(
+                client, primary_status, fallback_status
+            )
+        )
+
+    def ads_connections_status_check_result(
+        self,
+        client: framework.helpers.docker.Client,
+        primary_status: channelz_pb2.ChannelConnectivityState,
+        fallback_status: channelz_pb2.ChannelConnectivityState,
```
sergiitk, line 147: bad type hint

Image

```suggestion
        primary_status: channelz_pb2.ChannelConnectivityState.State,
        fallback_status: Optional[channelz_pb2.ChannelConnectivityState.State],
```

assert_ads_connections signature needs to be updated too

tests/fallback_test.py
```diff
@@ -143,11 +156,9 @@ def assert_ads_connections(
...
                     milliseconds=_STATUS_POLL_INTERVAL_MS.value
                 ),
-            ),
-            primary_status,
-        )
-        self.assertEqual(
-            client.expect_channel_status(
+            )
+            == primary_status
+            and client.expect_channel_status(
```
sergiitk, line 161: This is unreadable. Please break up into smaller cases similar to https://google.github.io/styleguide/pyguide.html#211-conditional-expressions

It should be something like

```py
    def ads_connections_status_check_result(
        self,
        client: framework.helpers.docker.Client,
        expected_primary_status: channelz_pb2.ChannelConnectivityState.State,
        expected_fallback_status: channelz_pb2.ChannelConnectivityState.State,
    ) -> bool:
        primary_status = client.expect_channel_status(
            self.bootstrap.primary_port,
            expected_primary_status,
            timeout=datetime.timedelta(milliseconds=_STATUS_TIMEOUT_MS.value),
            poll_interval=datetime.timedelta(
                milliseconds=_STATUS_POLL_INTERVAL_MS.value
            ),
        )

        fallback_status = client.expect_channel_status(
            self.bootstrap.fallback_port,
            expected_fallback_status,
            timeout=datetime.timedelta(milliseconds=_STATUS_TIMEOUT_MS.value),
            poll_interval=datetime.timedelta(
                milliseconds=_STATUS_POLL_INTERVAL_MS.value
            ),
        )
        return (
            primary_status == expected_primary_status
            and fallback_status == expected_fallback_status
        )
```

But check my code my code for correctness, I didn't test it, just moved things around.

tests/fallback_test.py
```diff
@@ -306,28 +317,51 @@ def test_fallback_mid_update(self):
...
+                    and "server3" in stats.rpcs_by_peer,
+                )
+                retryer(client.get_stats, 10)
+
+    def check_ads_connections_statuses(
+        self, client, primary_status, fallback_status
+    ):
+        retryer = retryers.constant_retryer(
+            wait_fixed=datetime.timedelta(seconds=1),
+            timeout=datetime.timedelta(seconds=60),
```
sergiitk, line 356: Please note that expect_channel_status already does some polling: https://github.com/grpc/psm-interop/blob/5465f4b7d675862fd253fa54f7ebc53d2e64d6c8/framework/helpers/docker.py#L303-L312

It polls every `0.3` seconds  (`_STATUS_POLL_INTERVAL_MS`) with total timeout `15` seconds (_STATUS_TIMEOUT_MS). And we have two calls to this method in the check function.

Meaning with this setup in case both call fail, we're not even going to have a chance to do the second retry. Let's bump it up a bit, something like


```suggestion
            timeout=datetime.timedelta(minutes=3),
```

tests/fallback_test.py
sergiitk: Yea, I see you pretty much copied the original signature. We should add a type checker CI to avoid problems like this in the first place.

sergiitk (approved):

tests/fallback_test.py
kannanjgithub: Done.

tests/fallback_test.py
kannanjgithub: Done.

PR #189, opened 2025-07-09
Disable Cloud run tests for Java as the feature needs more work.

kannanjgithub:


tests/fallback_test.py
```diff
@@ -14,6 +14,7 @@
 import datetime
 import logging
 import socket
+import time
```
ejona86, line 17: Delete?

kannanjgithub: Done.

ejona86 (approved):

PR #195, opened 2025-08-19
Fix typo in timeout value

kannanjgithub:
Follow-up to PR #193 to fix typo in timeout value. Also inlined some checks instead of creating separate functions for them.


ejona86 (approved):
The lambda formatting is quite awkward.

PR #198, opened 2025-08-26
Unset GKE zone/region for zonal/regional clusters

arjan-bal:
Passing Kokoro job: https://source.cloud.google.com/results/invocations/a579962a-4364-4412-ad96-cc5f0f01e412

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -849,8 +849,8 @@ psm::tools::log() {
 # Determines the cluster name and zone based on the given cluster identifier.
 # Globals:
 #   GKE_CLUSTER_NAME: Set to reflect the cluster name to use
-#   GKE_CLUSTER_ZONE: Set to reflect the cluster zone to use.
-#   GKE_CLUSTER_REGION: Set to reflect the cluster region to use (for regional clusters).
+#   GKE_CLUSTER_ZONE: Set to reflect the cluster zone to use. This variable will be unset for regional clusters.
+#   GKE_CLUSTER_REGION: Set to reflect the cluster region to use (for regional clusters). This variable will be unset for zonal clusters.
 # Arguments:
 #   The cluster identifier
 # Outputs:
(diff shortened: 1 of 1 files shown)


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -861,6 +863,7 @@ activate_gke_cluster() {
     GKE_CLUSTER_PSM_LB)
       GKE_CLUSTER_NAME="psm-interop-lb-primary"
       GKE_CLUSTER_ZONE="us-central1-a"
+      unset GKE_CLUSTER_REGION
```
sergiitk, line 866: What consumes `GKE_CLUSTER_REGION` and `GKE_CLUSTER_ZONE`?  Why not check if the variable is set to an empty string there?

arjan-bal: Moved the unset statements to the top of the function. I'm using unset instead of setting them to an empty string as it's generally safer when an empty string is a valid value.

Triggered a re-run of the cleanup job: https://source.cloud.google.com/results/invocations/a55ab8ce-7b36-4552-b06b-d8ce380694fa

> Also, I think we'll still need this change https://github.com/grpc/grpc/pull/40552.

No, I don't think that PR is required. It was a rollback PR for a required fix made in https://github.com/grpc/grpc/pull/40549. Even though fix was worked for the regional CSM test cluster, it caused problems in the cleanup of the dualstack cluster. I considered rolling back the fix earlier, but didn't because the impact was similar with/without the change. This PR fixes the cleanup of the dualstack cluster, so everything should work once this is submitted.

sergiitk:
Oh, the test run is executed against your branch in grpc/grpc:

```
github_scm_revision 	 {
  commit_sha: "psm-inteorp/fix-regional-cluster-cleanup"
  name: "grpc"
  owner: "arjan-bal"
  repository: "grpc"
}
```

Please run against master:


```
github_scm_revision 	 {
  commit_sha: "master"
  name: "grpc"
  owner: "grpc"
  repository: "grpc"
}
```

PR #203, opened 2025-09-12
Wait for namespace to become active

pawbhard:
Wait for namespace to become active

We are seeing many namespace not found error in internal bug b/258546394

[ ] [grpc/go/master/branch/xds_url_map](https://source.cloud.google.com/results/invocations/bb9efa36-ef9a-457a-b4cf-a80e3d911793)

### DIFF
--- a/framework/test_app/runners/k8s/k8s_base_runner.py
+++ b/framework/test_app/runners/k8s/k8s_base_runner.py
@@ -196,7 +196,7 @@ class KubernetesBaseRunner(base_runner.BaseRunner, metaclass=ABCMeta):
             self.namespace = self._create_namespace(
                 self.namespace_template, namespace_name=self.k8s_namespace.name
             )
-            if not self.k8s_namespace._namespace_active(self.namespace):  
+            if not self.k8s_namespace._namespace_active(self.namespace):
                 self.k8s_namespace.wait_for_namespace_active()
 
     def _start_completed(self):


arjan-bal:
We had a short discussion about this offline. If the k8s API server has returned a success response for the namespace creation request, it should not return a 404 when a deployment attempts to use it. This issue should probably be fixed by the GKE team. We should reach out to the GKE team for investigation. We can try to upgrade the GKE version for our cluster to see if it helps in the meantime.

framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -196,6 +196,7 @@ def run(self, **kwargs):
             self.namespace = self._create_namespace(
                 self.namespace_template, namespace_name=self.k8s_namespace.name
             )
+            self.k8s_namespace.wait_for_namespace_active()
```
sergiitk: We should be leave a log message when we're waiting for something so it doesn't look like the driver is just stuck doing nothing

```suggestion
            logger.info(
                "Waiting for namespace %s to become active",
                self.k8s_namespace.name,
            )
            self.k8s_namespace.wait_for_namespace_active()
```

framework/infrastructure/k8s.py
```diff
@@ -1286,3 +1286,36 @@ def _replicas_available(cls, deployment: V1Deployment, count: int) -> bool:
...
+        retryer = retryers.constant_retryer(
+            wait_fixed=_timedelta(seconds=wait_sec),
+            timeout=timeout,
+            check_result=self._namespace_active,
+        )
+        try:
+            retryer(self.get)
+        except retryers.RetryError as retry_err:
+            result = retry_err.result()
+            note = framework.errors.FrameworkError.note_blanket_error_info_below(
```
sergiitk: This method prints out the following

```
################################################################################
# IMPORTANT: This is not a root cause. This is an indication that
# _something_ -- literally _anything_ -- has gone wrong in the xDS flow.
# It is _your_ responsibility to look through the interop client and/or
# server logs to determine what exactly went wrong.
################################################################################
```

I don't think this has anything to do with the xds flow.

framework/infrastructure/k8s.py
```diff
@@ -1286,3 +1286,36 @@ def _replicas_available(cls, deployment: V1Deployment, count: int) -> bool:
...
+
+    def wait_for_namespace_active(
+        self,
+        timeout_sec: int = WAIT_SHORT_TIMEOUT_SEC,
+        wait_sec: int = WAIT_SHORT_SLEEP_SEC,
+    ) -> None:
+        timeout = _timedelta(seconds=timeout_sec)
+        retryer = retryers.constant_retryer(
+            wait_fixed=_timedelta(seconds=wait_sec),
+            timeout=timeout,
```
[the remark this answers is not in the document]
sergiitk: be careful - 4038975 applied this change incorrectly

framework/test_app/runners/k8s/k8s_base_runner.py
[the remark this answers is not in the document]
pawbhard: DOne

pawbhard:
latest run [grpc/go/master/branch/xds_url_map](https://source.cloud.google.com/results/invocations/a271b6aa-84db-48a6-a544-2ba4950fba13)

framework/infrastructure/k8s.py
```diff
@@ -1286,3 +1286,32 @@ def _replicas_available(cls, deployment: V1Deployment, count: int) -> bool:
...
+    def _namespace_active(cls, namespace: V1Namespace) -> bool:
+        return (
+            isinstance(namespace, V1Namespace)
+            and namespace.status.phase == "Active"
+        )
+
+    def wait_for_namespace_active(
+        self,
+        timeout: _timedelta = WAIT_SHORT_TIMEOUT_SEC,
+        retry_wait: _timedelta = WAIT_SHORT_SLEEP_SEC,
```
sergiitk: you've missed it again:

```
        timeout: _timedelta = WAIT_SHORT_TIMEOUT,
        retry_wait: _timedelta = WAIT_SHORT_SLEEP,
```

note there's no `_SEC`

sergiitk:
```
WAIT_SHORT_TIMEOUT: Final[_timedelta] = _timedelta(minutes=1)
WAIT_SHORT_TIMEOUT_SEC: Final[int] = WAIT_SHORT_TIMEOUT.seconds
```

`WAIT_SHORT_TIMEOUT` is timedelta
`WAIT_SHORT_TIMEOUT_SEC` is integer seconds

PR #236, opened 2026-03-24
Added retry for authz test to remove flakiness

pawbhard:
Authz test sometime fails usually in first or second step due to not receiving policy from TD.

This PR add a retry mechanism to account for this.
Internal bug b/331206277

Pass Run (first commit): https://source.cloud.google.com/results/invocations/7cab4574-e705-4aec-8cfb-15d41e03213e


tests/authz_test.py
```diff
@@ -214,7 +218,17 @@ def configure_and_assert(
         )
         # b/228743575 Python has as race. Give us time to fix it.
         stray_rpc_limit = 1 if self.lang_spec.client_lang == _Lang.PYTHON else 0
-        self.assertRpcStatusCodes(
+
```
eshitachandwani: I don't think we should add a retryer here, rather we should add another function `assertRpcStatusCodesWithRetry` [here](https://github.com/grpc/psm-interop/blob/5ff4026ca30795364460dbf0031e97c801959117/framework/xds_k8s_testcase.py#L437) where the function is defined so that it can be re-used if needed and follows similar convention as other functions.

eshitachandwani (commented):
Also , the test is failing , can you add a comment explaining if the failure if not related to this change or if it is expected to fail?

eshitachandwani (approved):

PR #238, opened 2026-03-24
Migrate GKE workload identity from per-test KSA to principal identifiers

kannanjgithub:
Skip creating workload identity binding for k8s service accounts in clusters with pre-existing bindings created for cluster level principal identifiers. Introduces support for a mode where Workload Identity is enabled, but the IAM binding is skipped. Adding a new flag: --workload_identity_iam_policy_binding (enabled by default to preserve current behavior). However, when this flag is disabled via --noworkload_identity_iam_policy_binding, the framework will skip the creation and management of the temporary KSA-to-GSA binding. Instead, it will assume that a permanent binding is already in place.

Created the workload identity binding using cluster principal identifiers, for impersonating the gcp service accounts in the prod clusters: psm-interop-lb-primary, psm-interop-lb-secondary, psm-interop-security and psm-interop-fleet-cluster.

LB, security and Spiffe tests use the default enable_workload_identity value of True, so I have configured these tests with --noworkload_identity_iam_policy_binding to skip creating the per-KSA binding to GSA so that the pre-existing cluster level binding applies. 
In the following ad-hoc test runs for these tests, the message ["Granting %s to %s for GCP Service Account %s"](https://github.com/grpc/psm-interop/blob/1b6b39bf8b1a8a96228a70ef0adf2217840bd36c/framework/test_app/runners/k8s/k8s_base_runner.py#L459) is not seen anymore since the flag --noworkload_identity_iam_policy_binding is passed now.
[LB tests](http://sponge2/a1ec5eaa-7e9f-4cd9-ac9b-c898a483630e)
[Security tests](https://btx.cloud.google.com/invocations/8b2884cf-ce3d-40fd-8200-fcc2dfae0ccf)
[Spiffe tests](https://btx.cloud.google.com/invocations/5b095e5e-ff93-4772-a86f-9f7636c17b7e)

The following tests pass --noworkload_identity_enabled flag and use the GCP compute engine service account directly so I have not modified their configurations for passing the workload_identity_iam_policy_binding flag, since the flag value won't be used in this case anyway.

Dual stack test was still [granting](https://screenshot.googleplex.com/BtSHJvz7yyqAt6b) workload identity at KSA level because of a missing parameter propagation in [xds_k8s_test_case.py](https://github.com/grpc/psm-interop/pull/238/changes#diff-0562d55945486687eb6462c4b7bc35be9eccc320294608618da8b3774a51b23d) in initKubernetesClientRunner and initKubernetesServerRunner for the enable_workload_identity flag value so it defaulted to True. I have now included both enable_workload_identity and the workload_identity_iam_policy_binding in these places, and now the per-KSA grant doesn't happen for [dual stack test run](https://btx.cloud.google.com/invocations/4a01e348-481a-466a-bd40-6b02306aef50).

[Url map tests](https://btx.cloud.google.com/invocations/97bcd12b-c44a-4cfc-b43b-85b4115876b1)
[Psm light tests](https://btx.cloud.google.com/invocations/d5bba10c-ab04-4f7c-a695-d0b65e0d2dc1)
[Gamma tests](https://btx.cloud.google.com/invocations/1e47ff4e-35ed-4d90-b683-3317b79dcf46)

[Cloud run tests](https://btx.cloud.google.com/invocations/c3cb9da1-0275-4050-8281-c3cb4d2e8078) don't pass --noworkload_identity_enabled flag but run on a non-workload identity enabled cluster, so there is no per-KSA grant.

b/475397692


pawbhard (approved):
LGTM

PR #243, opened 2026-04-06
Wait for client to be ready for cloud run testcases

pawbhard:
For cloud run we are not waiting for client to be ready, adding a wait for ready.

We have similar for other test cases.

This is causing 1 out of 100 RPC fail for cloud run test cases
Internal bug : b/432609370

### DIFF
--- a/framework/test_app/client_app.py
+++ b/framework/test_app/client_app.py
@@ -356,7 +356,7 @@ class XdsTestClient(framework.rpc.grpc.GrpcApp):
         )
         channel = retryer(
             self.find_active_xds_channel,
-            xds_server_uri,
+            xds_server_uri=xds_server_uri,
             rpc_deadline=rpc_deadline,
             secure_channel=secure_channel,
         )
--- a/framework/test_app/client_app.py
+++ b/framework/test_app/client_app.py
@@ -142,8 +142,11 @@ class XdsTestClient(framework.rpc.grpc.GrpcApp):
             log_target=f"{self.hostname}:{self.maintenance_port}",
         )
 
-    def get_csds_parsed(self, **kwargs) -> Optional[grpc_csds.DumpedXdsConfig]:
-        return self.csds.fetch_client_status_parsed(**kwargs)
+    def get_csds_parsed(
+        self, secure_channel: bool = False, **kwargs
+    ) -> Optional[grpc_csds.DumpedXdsConfig]:
+        csds: _CsdsClient = self.secure_csds if secure_channel else self.csds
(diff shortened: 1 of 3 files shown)


eshitachandwani:
Looks good to me, but can you add a test run for the cloud run test?

eshitachandwani:
Also can you fix the pylint error. You will need to change the function signature of [this function](https://github.com/grpc/psm-interop/blob/45da58197b8fec30d2e3f1894f43283349560af4/framework/test_cases/spiffe_testcase.py#L74) too , since this inherits and overrides the `startCloudRunClient` function from the base class.

pawbhard:
Run: http://sponge2/8e55d4db-f61f-44fe-8d1e-0038834e7504

pawbhard:
> Also can you fix the pylint error. You will need to change the function signature of [this function](https://github.com/grpc/psm-interop/blob/45da58197b8fec30d2e3f1894f43283349560af4/framework/test_cases/spiffe_testcase.py#L74) too , since this inherits and overrides the `startCloudRunClient` function from the base class.

Yes, done , fixed pylint and started test run.

pawbhard:
@eshitachandwani , Debugging this  I see this failing with timeout in getting ready. We might be encountering same TD issue. I want to check why sending RPC is causing it to work that is strange. Give me some time to analyse this.

pawbhard:
Latest run : http://sponge2/6fbd776c-c9a2-41b8-bf7c-356ecf08a67f

pawbhard:
> @eshitachandwani , Debugging this I see this failing with timeout in getting ready. We might be encountering same TD issue. I want to check why sending RPC is causing it to work that is strange. Give me some time to analyse this.

Oaky the issue seems to be for cloud run we have to use secure channel, wait for ready was using unsecure, causing the failure in the run. Added plumbing and secure_channel bool similar to other place for wait for ready. NOw test is parring . Added latest run.

PTAL @eshitachandwani

framework/test_app/client_app.py
```diff
@@ -342,8 +356,9 @@ def wait_for_xds_channel_active(
         )
         channel = retryer(
             self.find_active_xds_channel,
-            xds_server_uri=xds_server_uri,
```
eshitachandwani, line 345: This change seems unrelated, can we revert this?

eshitachandwani (approved):
LGTM except one comment.

framework/test_app/client_app.py
pawbhard: Reverted

PR #246, opened 2026-04-15
fix: prevent race conditions causing plain text communication

AgraVator:
Currently, a race condition exists where pods connect to Traffic Director before security policies are fully propagated, causing them to fallback to plaintext configurations. This leads to unrecoverable protocol mismatches (e.g., SSL wrong version number errors) in the initial attemps.

To eliminate test flakiness and avoid the need for the assertTestAppSecurityWithRetry retry loop, we must ensure that the client and server receive their intended TLS/mTLS security configurations in their very first xDS responses from Traffic Director.

[Security Test Run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fjava%2Fv1.80.x%2Fbranch%2Fpsm-security/activity/0735af33-fd5e-497c-8512-2fab3bd01530/log)
[Authz Test Run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fjava%2Fv1.81.x%2Fbranch%2Fpsm-security/activity/b99e8f34-4791-4c87-a071-ada9da6acc52/log)


pawbhard:
@eshitachandwani and @arjan-bal Please review the PR.
Abhishek, lets make sure to run securtity and authz test run with the changes.

arjan-bal:
Apologies for the delay, have been busy last week. Will leave a review within the next couple of days.

framework/infrastructure/traffic_director.py
```diff
@@ -1085,6 +1140,72 @@ def setup_client_security(
...
         )
 
+    @staticmethod
+    def wait_for_server_tls_ready(
+        test_server, server_port: int, timeout_sec: int = 120
+    ):
+        import datetime
+
+        from framework.helpers import retryers
+        from framework.rpc import grpc_csds
```
arjan-bal: style: Is there a reason to have these imports inside the method instead of at the top? The style guide recommends having all the imports at the top.

https://google.github.io/styleguide/pyguide.html#313-imports-formatting

framework/infrastructure/traffic_director.py
```diff
@@ -1085,6 +1140,72 @@ def setup_client_security(
...
+        channel = test_server._make_channel(test_server.maintenance_port)
+        csds_client = grpc_csds.CsdsClient(
+            channel,
+            log_target=f"{test_server.hostname}:{test_server.maintenance_port}",
+        )
+
+        def _check_tls_ready():
+            config = csds_client.fetch_client_status_parsed()
+            if not config:
+                logger.debug("CSDS config empty")
```
arjan-bal: Is this case possible without a bug? If yes, we should probably throw an exception instead. There are occcurances of the following in the repo: `self.assertIsNotNone(parsed)`

framework/infrastructure/traffic_director.py
```diff
@@ -1085,6 +1140,72 @@ def setup_client_security(
...
+                "DownstreamTlsContext" in lds_str or "require_tls" in lds_str
+            ):
+                logger.info("Server reported TLS readiness")
+                return True
+
+            logger.debug("LDS does not contain TLS config yet: %s", lds_str)
+            return False
+
+        retryer = retryers.constant_retryer(
+            wait_fixed=datetime.timedelta(seconds=2),
```
arjan-bal: nit:  Is there a reason for choosing 2 seconds? It seems to frequent and may lead to log spam. Maybe we should increase it to more than 10 seconds?

framework/infrastructure/traffic_director.py
```diff
@@ -1085,6 +1140,72 @@ def setup_client_security(
...
+            try:
+                config = csds_client.fetch_client_status_parsed()
+                if config and config.lds:
+                    logger.error(
+                        "Last received LDS config: %s", str(config.lds)
+                    )
+                else:
+                    logger.error(
+                        "Last received CSDS config was empty or lacked LDS"
+                    )
```
arjan-bal: nit: Is there any harm in printing the entire config? That would simplify the code and avoid the ambiguous "config was empty or lacked LDS" case.

framework/infrastructure/traffic_director.py
```diff
@@ -1067,6 +1075,53 @@ def __init__(
         self.authz_policy: Optional[AuthorizationPolicy] = None
         self.endpoint_policy: Optional[EndpointPolicy] = None
 
+    def setup_for_grpc_with_security(
```
arjan-bal: It appears that `setup_for_grpc_with_security` is implemented entirely by composing public methods and there is only one caller. Given this, could we inline this function directly into its caller, `setupTrafficDirectorGrpcWithSecurity`? This would avoid having methods that merely delegate to other methods, thereby keeping the `TrafficDirectorSecureManager` API simpler.

See https://refactoring.com/catalog/inlineFunction.html

framework/infrastructure/traffic_director.py
AgraVator: If there is no xDS client initialized or active in the application:

The CSDS RPC itself will still succeed: Assuming the application is compiled with CSDS support and the service is exposed on the port we are querying, the server will respond to the ClientStatusDiscoveryService.FetchClientStatus request without an RPC error.
The response content will be empty: The ClientStatusResponse message contains a list of ClientConfig messages. If there are no active xDS clients, this list will be empty.

AgraVator:
[Latest Test Run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fjava%2Fv1.80.x%2Fbranch%2Fpsm-security/activity/ce9b0363-fe66-4434-a702-9acc6a8b7a6d/log)

framework/infrastructure/traffic_director.py
```diff
@@ -1085,6 +1096,60 @@ def setup_client_security(
             server_namespace, server_name
         )
 
+    @staticmethod
+    def wait_for_server_tls_ready(
+        test_server, server_port: int, timeout_sec: int = 120
+    ):
```
arjan-bal: As discussed offline, the test shouldn't need to wait for the server to receive the TLS policy. Because the server's TLS config is established at the beginning rather than applied through a patch, we can safely assume both parties are using TLS as long as the test waits for the client to have a ready server channel.

arjan-bal:
@pawbhard should we remove the retires introduced in https://github.com/grpc/psm-interop/pull/237?

pawbhard:
> @pawbhard should we remove the retires introduced in #237?

Will be done after this PR

eshitachandwani:
The changes look good to me. But since we are adding a field in the function that is used by almost all the tests , can you please run one more test, other than security to make sure it is not failing anything else. Since it is an optional field, it should not affect other tests, but we should still verify.
Also please fix the pylint.

AgraVator:
[Latest security test run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fjava%2Fv1.80.x%2Fbranch%2Fpsm-security/activity/486d90cd-1911-41ee-9cd9-2dc876974ad3/log)
[Latest random test run](https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fnode%2Fv1.14.x%2Fxds_k8s_lb/activity/ed45e8f6-65e4-4bdb-a7f7-16744a0f7666/log)

arjan-bal (approved):
LGTM!

PR #252, opened 2026-05-28
[Framework] Add request IDs to POST and PATCH operations for idempotency

murgatroid99:
This is an attempt to fix a bug (internal reference b/280522546) where a POST or PATCH request will succeed on the server but the client will not receive the success response, and then subsequent retries will fail because the resource already exists, or has already been modified.

The attempted fix is to add a `request_id` parameter to those requests, so that retries will hopefully succeed. The [request ID documentation](https://docs.cloud.google.com/compute/docs/reference/rest/v1/urlMaps/insert#query-parameters) describes the parameter like this:

> An optional request ID to identify requests. Specify a unique request ID so that if you must retry your request, the server will know to ignore the request if it has already been completed.

### DIFF
--- a/framework/infrastructure/gcp/compute.py
+++ b/framework/infrastructure/gcp/compute.py
@@ -661,21 +661,21 @@ class ComputeV1(
         logger.info(
             "Creating compute resource:\n%s", self.resource_pretty_format(body)
         )
-        request_id = uuid.uuid4()
+        request_id = str(uuid.uuid4())
         if region:
             resp = self._execute(
                 collection.insert(
                     project=self.project,
                     region=region,
                     body=body,
-                    request_id=request_id,
+                    requestId=request_id,
                 ),
                 region=region,
             )
         else:
             resp = self._execute(
                 collection.insert(
-                    project=self.project, body=body, request_id=request_id
+                    project=self.project, body=body, requestId=request_id
                 )
             )
         return self.GcpResource(body["name"], resp["targetLink"])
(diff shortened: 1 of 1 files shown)


murgatroid99:
- [x] [grpc/java/v1.81.x/branch/psm-security](https://source.cloud.google.com/results/invocations/4c072b36-a51e-433e-a52e-7602a429f22e)
- [x] [grpc/java/v1.81.x/branch/xds_k8s_lb](https://source.cloud.google.com/results/invocations/7a75fa20-5de6-4c2a-93fa-eceba0254628)
- [x] [grpc/java/v1.81.x/branch/xds_url_map](https://source.cloud.google.com/results/invocations/5635c09f-79e8-489f-a29f-aef662327c8d)

framework/infrastructure/gcp/compute.py
```diff
@@ -660,25 +661,34 @@ def _insert_resource(
...
                 ),
                 region=region,
             )
         else:
             resp = self._execute(
-                collection.insert(project=self.project, body=body)
+                collection.insert(
+                    project=self.project, body=body, request_id=request_id
+                )
             )
```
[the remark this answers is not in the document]
sergiitk: This is a pretty good catch. I'm impressed.

framework/infrastructure/gcp/compute.py
sergiitk, line 710: Do we want to add requestId on delete operation too? It should be supported; example: https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.healthChecks.html#delete

sergiitk (approved):
LGTM, with a comment

framework/infrastructure/gcp/compute.py
murgatroid99: I don't think it matters, because an error wouldn't make the test fail. It just gets logged and swallowed.

pawbhard (approved):
LGTM, Good catch

pawbhard (approved):
LGTM, Good catch

PR #253, opened 2026-06-01
add regional td test and setup kokoro script

shivaspeaks:
go/regional-td-grpc-test-plan


framework/infrastructure/gcp/compute.py
```diff
@@ -190,37 +191,69 @@ def create_backend_service_traffic_director(
             body["localityLbPolicies"] = locality_lb_policies
         if outlier_detection:
             body["outlierDetection"] = outlier_detection
+        if region:
```
kannanjgithub: There is no need to have this branching for if region as self._insert_resource handles None value for region, so you can unconditionally pass region=region here. Same holds for all other helper methods.

shivaspeaks: Discussed offline. Since global and regional apis are also different for both branches, I'll go ahead and make the return inline with conditional.

framework/xds_k8s_testcase.py
```diff
@@ -1133,10 +1134,22 @@ def _start_test_client(
...
+                    timeout=wait_for_active_ads_timeout,
+                )
+            except retryers.RetryError:
+                if getattr(self, "xds_server_region", None):
+                    test_client.wait_for_active_xds_channel(
+                        xds_server_uri=(
+                            f"trafficdirector.{self.xds_server_region}"
+                            ".rep.googleapis.com:443"
+                        ),
+                        timeout=wait_for_active_ads_timeout,
```
kannanjgithub: Every time a regional test starts, the framework incurs a `wait_for_active_ads_timeout` timeout penalty attempting to connect to the global Traffic Director when it has no connection to the global control plane, before finally falling back to checking the regional endpoint. 
Optimize this.

tests/regional_traffic_director_test.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+
+        with self.subTest("04_start_test_servers"):
+            test_servers = self.startTestServers()
+
+        with self.subTest("05_add_server_backends_to_backend_services"):
+            self.setupServerBackends()
+
+        with self.subTest("06_start_test_client"):
+            if self.xds_server_region:
```
pawbhard: Use existing test.py file, we will run a new test but will use same py file

pawbhard (changes requested):

framework/infrastructure/traffic_director.py
```diff
@@ -950,12 +971,107 @@ def __init__(
         self.http_route: Optional[HttpRoute] = None
         self.mesh: Optional[Mesh] = None
 
+    def create_backend_service(
```
pawbhard: Why we needed this new function, how it is done in current state

pawbhard (changes requested):
We are doing if else for region for the data member which is same we don't need seprate for regional as one tetscase will not have regional and non regional. Only thing we need to do is while creating use the flag and store and delete don't need to change

framework/infrastructure/traffic_director.py
shivaspeaks: Simplified the usage.

framework/infrastructure/gcp/compute.py
```diff
@@ -190,18 +191,37 @@ def create_backend_service_traffic_director(
...
+            region=region,
+        )
 
-    def get_backend_service_traffic_director(self, name: str) -> "GcpResource":
+    def get_backend_service_traffic_director(
+        self, name: str, region: Optional[str] = None
+    ) -> "GcpResource":
         return self._get_resource(
-            self.api.backendServices(), backendService=name
+            self.api.regionBackendServices()
```
pawbhard, line 210: this seems like an anti pattern to me we are using region in if statement here and also passing region down to the function which aagin have the if statement. Please find a way to make this cleaner and simpler. @sergiitk can help with the patterns to use in python.

same for other places as well

pawbhard (dismissed):

framework/infrastructure/gcp/compute.py
shivaspeaks: Passing down the explicit regional vs global collections while also passing the region keyword arg into `_insert_resource` resulted in messy inline if branches everywhere. I've done a major refactor of `compute.py` to make this much cleaner in my local:

1. I added a private helper `_get_collection(self, collection_name: str, region: Optional[str])` to completely encapsulate the SDK mapping logic between global and regional API property names (e.g. `backendServices()` vs `regionBackendServices()`).
2. I refactored all base compute methods (`_insert_resource`, `_get_resource`, `_patch_resource`, `_delete_resource`, `_exists_resource`) to accept a simple string `collection_name` instead of a pre-resolved `googleapiclient` collection object.
3. Finally, I went through every public helper method in `compute.py` (like `create_target_grpc_proxy`, `create_backend_service_traffic_director`, `create_url_map`, etc.) and removed the inline if region else branching. They now cleanly pass the string identifier (e.g. "targetGrpcProxies") straight down to the base methods.

Now, the public wrappers are extremely flat and simple, and the base methods dynamically resolve the correct SDK endpoint under the hood using region. No more duplicated if-statements! I will go ahead and make a commit if you're fine with this way. 

Note: I was not being invasive with the existing flow of things and just creating a new branch using if else such that it won't affect existing way the tests work.

config/common-regional_td.cfg
```diff
@@ -0,0 +1,3 @@
+--flagfile=config/common.cfg
+--xds_server_region=us-west1
+--noworkload_identity_iam_policy_binding
```
sergiitk: Missing a newline at the end of the file.
You can configure VS Code to add it automatically by setting

```json5
"files.insertFinalNewline": true,
```


```suggestion
--noworkload_identity_iam_policy_binding

```

framework/infrastructure/gcp/compute.py
sergiitk: > I've done a major refactor of compute.py to make this much cleaner in my local
> ...

Yes, this refactoring makes sense to me. Let me know when it'll be ready, I can take a look

sergiitk: Also feel free to remove branching from `def _insert_resource`. I'm pretty sure everything will work with one call

```diff
diff --git a/framework/infrastructure/gcp/compute.py b/framework/infrastructure/gcp/compute.py
index 1a3cbeb802..4c4128e34c 100644
--- a/framework/infrastructure/gcp/compute.py
+++ b/framework/infrastructure/gcp/compute.py
@@ -662,22 +662,15 @@ class ComputeV1(
             "Creating compute resource:\n%s", self.resource_pretty_format(body)
         )
         request_id = str(uuid.uuid4())
-        if region:
         resp = self._execute(
             collection.insert(
                 project=self.project,
                 region=region,
                 body=body,
                 requestId=request_id,
             ),
             region=region,
         )
-        else:
-            resp = self._execute(
-                collection.insert(
-                    project=self.project, body=body, requestId=request_id
-                )
-            )
         return self.GcpResource(body["name"], resp["targetLink"])
 
     def _patch_resource(self, collection, body, **kwargs):
```

shivaspeaks:
The comments on network_services.py are invalid because our internal `self.parent(location)` helper explicitly catches location is None and defaults it to "global", so there's no risk of an invalid resource path.
https://github.com/grpc/psm-interop/blob/7ab712055f1c94d5e172608e6912b708b530bd65/framework/infrastructure/gcp/api.py#L536-L539

framework/xds_k8s_testcase.py
```diff
@@ -1133,8 +1134,16 @@ def _start_test_client(
             server_target=server_target, **kwargs
         )
         if wait_for_active_ads:
+            if self.xds_server_region:
+                expected_uri = (
+                    f"trafficdirector.{self.xds_server_region}"
+                    ".rep.googleapis.com:443"
+                )
```
kannanjgithub: The code introduces string templates to build critical URLs and hosts (trafficdirector.{region}.rep.googleapis.com:443 and xds://traffic-director.{region}.xds.googleapis.com/...), but no unit tests were added to guard this logic against typos or regressions. Add a simple standalone unit test like https://paste.googleplex.com/4797133810171904

kannanjgithub (approved):

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -895,6 +917,10 @@ activate_gke_cluster() {
       GKE_CLUSTER_NAME="psm-interop-fleet-cluster"
       GKE_CLUSTER_ZONE="us-central1-a"
       ;;
+    GKE_CLUSTER_PSM_REGIONAL_TD)
+      GKE_CLUSTER_NAME="psm-interop-lb-secondary"
+      GKE_CLUSTER_ZONE="us-west1-b"
```
sergiitk, line 922: I wonder why we're using secondary cluster in a primary role. Was there anything special about its setup? Can we use `GKE_CLUSTER_PSM_LB` or `GKE_CLUSTER_PSM_BASIC` instead?

framework/infrastructure/gcp/compute.py
```diff
@@ -618,77 +693,123 @@ def delete_neg_serverless(self, name: str, region: str):
...
+                return self.api.globalForwardingRules()
+            elif collection_name == "operations":
+                return self.api.globalOperations()
+            else:
+                return getattr(self.api, collection_name)()
+
     def _insert_resource(
         self,
-        collection: discovery.Resource,
+        collection_name: str,
```
sergiitk: I disagree with this approach. Please change it back to `collection: discovery.Resource`.

Passing collection name here just to send it to a switch function that returns an actual object makes things very obscure. Example: `forwardingRules` collection is not the same as `globalForwardingRules`, but in the create function it looks like we always trying to create a regional one.

```py
        return self._insert_resource(
            "forwardingRules",
            body,
            region=region,
        )
```

framework/infrastructure/traffic_director.py
```diff
@@ -257,7 +259,13 @@ def create_backend_service(
             protocol = _BackendGRPC
 
         name = self.make_resource_name(self.BACKEND_SERVICE_NAME)
-        logger.info('Creating %s Backend Service "%s"', protocol.name, name)
+        region = self.region
```
sergiitk: nit: use `self.region` without re-assigning to improve code readability.

config/common-regional_td.cfg
sergiitk, line 1: rename to `common-regional-td.cfg` for consistency

framework/xds_k8s_testcase.py
```diff
@@ -1133,8 +1134,16 @@ def _start_test_client(
             server_target=server_target, **kwargs
         )
         if wait_for_active_ads:
+            if self.xds_server_region:
+                expected_uri = (
+                    f"trafficdirector.{self.xds_server_region}"
+                    ".rep.googleapis.com:443"
```
sergiitk: No, we can't assume that this is the correct Control Plane URI in all cases just because a region is set. This simply won't work in staging, with TD sandboxes or our mock control plane.
We need configure the correct Control Plane URI (`--xds_server_uri`) explicitly `common-regional-td.cfg`

framework/xds_k8s_testcase.py
```diff
@@ -1232,7 +1244,14 @@ def startTestServers(
     def startTestClient(
         self, test_server: XdsTestServer, **kwargs
     ) -> XdsTestClient:
-        return self._start_test_client(test_server.xds_uri, **kwargs)
+        if self.xds_server_region:
+            server_target = (
+                f"xds://traffic-director.{self.xds_server_region}"
+                f".xds.googleapis.com/{test_server.xds_address}"
+            )
```
sergiitk: We need to configure `test_server.xds_uri` / `test_server.xds_address` correctly and we won't need any custom uri generation that only supports one endpoint and will fail in staging. As far as I understand, we'll need to add `xds_authority` field and handle same as `xds_host`/`xds_port`. See https://github.com/grpc/psm-interop/blob/7ab712055f1c94d5e172608e6912b708b530bd65/framework/test_app/server_app.py#L133-L146

sergiitk (changes requested):

.kokoro/psm_interop_kokoro_lib.sh
shivaspeaks: This was agreed in the design. We were mainly avoiding us-central-1 zone. 

Besides, I believe "secondary" is just the name, is there anything special will secondary?

config/common-regional_td.cfg
shivaspeaks: Should we? We are naming the suite as reginal_td(check the BUILD file in cl/923271005) and this comes from there for consistency, using "regional_td" as a word together. For example cfgs in language repos also have same name: 
java: https://github.com/grpc/grpc-java/blob/master/buildscripts/kokoro/psm-regional_td.cfg
c++: https://github.com/grpc/grpc/blob/master/tools/internal_ci/linux/psm-regional_td.cfg
go: https://github.com/grpc/grpc-go/blob/master/test/kokoro/psm-regional_td.cfg

I can go ahead and change. Just wanted to point all this out.

framework/infrastructure/gcp/compute.py
shivaspeaks: Yes, I agree with you. Previous state was more readable. Even I was not satisfied when committing this change.

framework/xds_k8s_testcase.py
shivaspeaks: Hardcoded the URI in the config file and it decoupled the GCP Region from the Control Plane URI.

config/common-regional_td.cfg
sergiitk: It's ok. If it's already like that in a different repo, I have no issues

.kokoro/psm_interop_kokoro_lib.sh
sergiitk: There is something special about it: https://github.com/grpc/psm-interop/blob/7ab712055f1c94d5e172608e6912b708b530bd65/.kokoro/psm_interop_kokoro_lib.sh#L922-L935

I'm ok with using it as a primary however. With one note: we need to change the "alias" from `GKE_CLUSTER_PSM_REGIONAL_TD` to `GKE_CLUSTER_PSM_LB_SECONDARY` to make it clear. These should match cluster name to avoid ambiguity.

sergiitk: In addition `GKE_CLUSTER_PSM_REGIONAL_TD` may give the impression it's a regional cluster (which is a thing too)

framework/infrastructure/gcp/compute.py
sergiitk: Glad to hear it!

framework/infrastructure/gcp/compute.py
```diff
@@ -265,79 +307,121 @@ def create_url_map(
...
         return self._insert_resource(
-            self.api.targetGrpcProxies(),
-            {
-                "name": name,
-                "url_map": url_map.url,
-                "validate_for_proxyless": validate_for_proxyless,
-            },
+            self.api.regionTargetGrpcProxies()
+            if region
+            else self.api.targetGrpcProxies(),
```
sergiitk: I'd probably move these above to be something like 

```py
        collection = (
            self.api.regionTargetGrpcProxies()
            if region
            else self.api.targetGrpcProxies()
        )
        return self._insert_resource(collection, body, region=region)
```

framework/infrastructure/gcp/compute.py
```diff
@@ -265,79 +307,121 @@ def create_url_map(
...
+            name,
+            region=region,
+        )
 
     def create_target_grpc_proxy(
         self,
         name: str,
         url_map: "GcpResource",
         validate_for_proxyless: bool = True,
+        region: Optional[str] = None,
```
sergiitk, line 391: nit: I see you already providing `region` as a kwarg everywhere, but there's also a good way to enforce it.
```suggestion
        *,
        region: Optional[str] = None,
```

framework/infrastructure/gcp/compute.py
```diff
@@ -772,14 +900,14 @@ def _wait(
 
         # TODO(sergiitk) try using wait() here
         # https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.globalOperations.html#wait
+        kwargs = {"project": self.project, "operation": operation_id}
```
sergiitk: kwargs is generally reserved to special `**kwargs` method parameter. let's use a different name and simplify this

```py
        request_args = {"project": self.project, "operation": operation_id}
        if region:
            request_args["region"] = region

        collection = (
            self.api.regionOperations()
            if region
            else self.api.globalOperations()
        )
        # TODO(sergiitk) try using wait() here
        # https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.globalOperations.html#wait
        op_request = collection.get(**request_args)
```

framework/infrastructure/gcp/compute.py
sergiitk:
```suggestion
        region: Optional[str] = None,
```

Not your PR, but we missed it

framework/test_app/server_app.py
```diff
@@ -142,6 +150,8 @@ def xds_address(self) -> str:
     def xds_uri(self) -> str:
         if not self.xds_host:
             return ""
+        if hasattr(self, "xds_authority") and self.xds_authority:
```
sergiitk: define it as a class field with setting the default

```py
class XdsTestServer(framework.rpc.grpc.GrpcApp):
  ...
  xds_authority: Optional[str] = None
```

then do `if self.xds_authority:` here

tests/unit/test_cases/regional_uri_test.py
```diff
@@ -0,0 +1,60 @@
...
+# limitations under the License.
+from unittest import mock
+
+from absl.testing import absltest
+
+# Since we want to test formatting logic residing in the test case base class:
+from framework.xds_k8s_testcase import RegularXdsKubernetesTestCase
+
+
+class RegionalUriTest(absltest.TestCase):
```
sergiitk, line 22: Appreciate the unit test!

sergiitk (approved):

PR #254, opened 2026-06-12
framework: add retries to GCP API get operations

easwars:
Adds `num_retries=self._GCP_API_RETRIES` to all `execute()` calls in `_get_resource` and health check paths. This helps mitigate HTTP 429 (Concurrent Operations Limit) errors from the Network Services API.

### DIFF
--- a/framework/infrastructure/gcp/api.py
+++ b/framework/infrastructure/gcp/api.py
@@ -572,7 +572,9 @@ class GcpStandardCloudApiResource(GcpProjectApiResource, metaclass=abc.ABCMeta):
         raise NotImplementedError
 
     def _get_resource(self, collection: discovery.Resource, full_name):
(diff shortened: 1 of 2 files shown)


pawbhard (approved):
LGTM, Please add a test pass run with the changes before merge

kannanjgithub:
In the failure I observed fusion2/f67a9556-0e39-4ee2-a79b-7d9573757673 in the security test failure, the 429 error is on health check patch operation of backend services. 
This PR does not really address that. 
And to deal with 429 error, adding num_retries is not the solution since the GCP API library will immediately retry the operation for the given number of times, and only exacerbate the problem even further. We will need a back off based retry for these operations.
If you are fine with it, I can raise a PR for that.

easwars:
Sure

PR #255, opened 2026-06-13
Revert "add regional td test and setup kokoro script (#253)"

pawbhard:
This reverts commit 8be128fd2cc48cc54a46fbbc37ea2a37a04e2cee.

This is causing regression in cloud run testcases causing continuous failures.

Passed Run : [9c2e13f0-98cc-46e9-ac37-2c1d3c351ff3](https://source.cloud.google.com/results/invocations/9c2e13f0-98cc-46e9-ac37-2c1d3c351ff3)

PR #256, opened 2026-06-15
fix cloud run test

shivaspeaks:
In `CloudRunMeshManager`, tracking the region for the Serverless NEG was accidentally done by setting `self.region = region`. However, because `CloudRunMeshManager` inherits from `TrafficDirectorManager`, this shadowed the base class's global `self.region` property. This inadvertently forced all Cloud Run tests (inbound, outbound, and spiffe) to behave as regional tests, which caused them to attempt to deploy regional Traffic Director components (backend services, routes) instead of global ones, leading to 404 Not Found API errors.
Fixed it by renaming `self.region` to `self.serverless_region` in `CloudRunMeshManager` to avoid scope collision with the base class, ensuring TD resources deploy globally while the Serverless NEGs deploy regionally.

There was one more issue, which was caught during deleting the resources. The Regional TD PR introduced a conditional branch in `compute.py` to route calls to `self.api.regionTargetGrpcProxies()` if a region was provided. However, GCP `TargetGrpcProxy` resources are strictly global objects and the API SDK has no such regional attribute. This caused regional tests to crash with an `AttributeError` during the `delete_target_grpc_proxy` teardown phase. Because the teardown crashed midway, it abandoned the rest of the deletion sequence, leaving dangling health checks and URL maps in the test project and causing 409 Already Exists conflicts on subsequent runs. Fix: Removed the regional logic and region arguments from `create_target_grpc_proxy` and `delete_target_grpc_proxy` in `compute.py`, guaranteeing they always route to the global `targetGrpcProxies()` API. Updated the `traffic_director.py` dispatcher to stop passing the region argument when provisioning gRPC proxies.

### DIFF
--- a/framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
+++ b/framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
@@ -50,7 +50,7 @@ class CloudRunMeshManager(td_base.TrafficDirectorAppNetManager):
         )
 
         # Settings
-        self.region = region
+        self.serverless_region = region
 
         # Managed resources
         self.neg: Optional[GcpResource] = None
@@ -90,10 +90,10 @@ class CloudRunMeshManager(td_base.TrafficDirectorAppNetManager):
         name = self.make_resource_name(self.NEG_NAME)
         logger.info("Creating serverless NEG %s", name)
         neg = self.compute.create_neg_serverless(
-            name, self.region, service_name
+            name, self.serverless_region, service_name
         )
         logger.info("Loading NEG %s", neg)
-        self.neg = self.compute.get_neg_serverless(name, self.region)
+        self.neg = self.compute.get_neg_serverless(name, self.serverless_region)
         return neg
 
     def delete_neg_serverless(self, force=False):
@@ -104,7 +104,7 @@ class CloudRunMeshManager(td_base.TrafficDirectorAppNetManager):
         else:
             return
         logger.info("Deleting serverless NEG %s", name)
-        self.compute.delete_neg_serverless(name, self.region)
+        self.compute.delete_neg_serverless(name, self.serverless_region)
         self.neg = None
 
     def cleanup(self, *, force=False):
--- a/framework/infrastructure/traffic_director.py
+++ b/framework/infrastructure/traffic_director.py
@@ -659,9 +659,12 @@ class TrafficDirectorManager:  # pylint: disable=too-many-public-methods
             target_proxy_type,
             self.url_map.name,
         )
-        self.target_proxy = create_proxy_fn(
-            name, self.url_map, region=self.region
-        )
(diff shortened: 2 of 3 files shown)


shivaspeaks:
Note: Please wait for me to run the test suite.

shivaspeaks:
test run: https://source.cloud.google.com/results/invocations/cfadb0a8-370b-4acd-bac2-4ace46f416ac

kannanjgithub (approved):

framework/infrastructure/mesh_resource_manager/cloud_run_mesh_manager.py
```diff
@@ -50,7 +50,7 @@ def __init__(
         )
 
         # Settings
-        self.region = region
+        self.serverless_region = region
```
pawbhard, line 53: Why is this name was causing issue, all I see is rename of local class member

framework/infrastructure/traffic_director.py
```diff
@@ -659,9 +659,12 @@ def create_target_proxy(self):
...
             self.url_map.name,
         )
-        self.target_proxy = create_proxy_fn(
-            name, self.url_map, region=self.region
-        )
+        if self.target_proxy_is_http:
+            self.target_proxy = create_proxy_fn(
+                name, self.url_map, region=self.region
+            )
+        else:
```
pawbhard, line 666: Why this if else, I think this will impact our ability to run the else case with regional TD in future

pawbhard (commented):
Fix seems to be just ignore the cloud run case. I think we agreed to have changes in a way that we would be able to run any testcase if needed with regional TD. (See if we can fix it correctly in follow up PR )

Also as mentioned in comment "https://github.com/grpc/psm-interop/pull/255#issuecomment-4699087544"

```
Please do run cloudrun one other non cloud run testcases before merging changes as changes are in framework.
```
@shivaspeaks  Always make sure to run multiple cases when changes are in framework, we missed this first time which have caused this issue and we are doing same again in this PR.

Example is running cloud run don't cover both cases of branching added 
"if self.target_proxy_is_http:"

@kannanjgithub please be mindful of this when approving PR.

kannanjgithub:
> @shivaspeaks Always make sure to run multiple cases when changes are in framework, we missed this first time which have caused this issue and we are doing same again in this PR.
> 
> Example is running cloud run don't cover both cases of branching added "if self.target_proxy_is_http:"
> 
> @kannanjgithub please be mindful of this when approving PR.

Cloud run inbound test uses target proxy with HTTP2 and cloud run outbound test uses target proxy with GRPC, so both cases were covered in the test run.

framework/infrastructure/traffic_director.py
kannanjgithub: GRPC backend services don't have regional, they are strictly global, hence this change.

PR #260, opened 2026-06-19
infra: Use exponential backoff for GCP API retries

kannanjgithub:
Refactor GCP API retry logic to use a centralized exponential backoff strategy instead of aggressive internal library retries. This addresses 429 (Too Many Requests) and 500 (Internal Error) failures observed in dense tests like security_test.py.

- Add exponential_retryer_with_attempts to framework/helpers/retryers.py.
- Update GcpProjectApiResource and ComputeV1 to use the new smart retryer for mutations, lookups, and listings.
- Disable internal library retries (num_retries=0) to give full control to the smart retryer and prevent "double spamming" the API.
- Fix missing retry logic in _get_resource lookups.

This approach ensures that when rate limits or transient flakes occur, the driver waits and backs off before retrying, providing a more resilient integration with GCP services.

Metabug b/522968808

[ad-hoc test run](http://shortn/_nehboLtOPq)

### DIFF
--- a/framework/helpers/retryers.py
+++ b/framework/helpers/retryers.py
@@ -91,6 +91,44 @@ def exponential_retryer_with_timeout(
     )
 
 
+def exponential_retryer_with_attempts(
+    *,
+    wait_min: timedelta,
+    wait_max: timedelta,
+    attempts: int,
+    retry_on_exceptions: Optional[_ExceptionClasses] = None,
+    retry_on_predicate: Optional[Callable[[BaseException], bool]] = None,
+    logger: Optional[logging.Logger] = None,
+    log_level: Optional[int] = logging.DEBUG,
+) -> Retrying:
+    if logger is None:
+        logger = retryers_logger
+    if log_level is None:
+        log_level = logging.DEBUG
+
+    retry_conditions = []
+    if retry_on_exceptions:
+        retry_conditions.append(
+            tenacity.retry_if_exception_type(retry_on_exceptions)
+        )
+    if retry_on_predicate:
+        retry_conditions.append(tenacity.retry_if_exception(retry_on_predicate))
+
+    if not retry_conditions:
+        retry_conditions.append(tenacity.retry_if_exception_type(Exception))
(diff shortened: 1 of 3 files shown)


eshitachandwani:
The test attached seems to be a failed run, can you please attach a successful run ?

eshitachandwani:
We seem to have code based retries defined [here](https://github.com/grpc/psm-interop/blob/04976d13e778c84a45b7448e873708fcfe812122/framework/infrastructure/k8s.py#L547) already. Can we try to use these?

kannanjgithub:
> We seem to have code based retries defined [here](https://github.com/grpc/psm-interop/blob/04976d13e778c84a45b7448e873708fcfe812122/framework/infrastructure/k8s.py#L547) already. Can we try to use these?

  1. Architectural Layering
   * Dependency Direction: framework/infrastructure/gcp/api.py is a foundational base class. Having it depend on framework/infrastructure/k8s.py (a specific implementation) would create a circular dependency or a "leaky abstraction."
   * Domain Specificity: The k8s retryer is specifically tuned for Kubernetes API behavior (e.g., waiting at least 10 seconds), which differs significantly from GCP API rate-limiting patterns.

  2. Technical Incompatibility
   * Exceptions: The k8s retryer catches client.ApiException and urllib3.exceptions.HTTPError. GCP API calls raise googleapiclient.errors.HttpError.
   * Wait Strategy: The k8s version has a minimum wait of 10 seconds. For GCP API 429s, a 1-second initial backoff is usually sufficient and much more efficient for CI test execution.
   * Retry Criteria: The k8s version uses a timeout-based stop (3 minutes), whereas the GCP implementation uses an attempt-based stop (5 attempts) to remain consistent with the existing _GCP_API_RETRIES flag logic.

kannanjgithub:
> The test attached seems to be a failed run, can you please attach a successful run ?

http://shortn/_LOISP0BLRj

eshitachandwani (approved):

PR #263, opened 2026-07-13
kokoro: Execute test cases concurrently

arjan-bal:
Related Issue: b/526886100

---

### Context & Summary
This change enables concurrent execution of test cases within the Security, LB, and DualStack test suites. Shifting from sequential execution to a parallelized model drastically reduces overall execution times and resolves the timeouts observed in our scheduled test runs.

The CSM suite remains unaffected and will continue to execute sequentially to prevent any shared resource conflicts.

---

### Performance Improvements

| Test Suite | Concurrent jobs | Before (mins) | After (mins) | Delta / Reduction (%) |
|---|---|---|---|---|
| Security | 2 | 220 | 122 | -44.5% |
| LB | 2 | 281 | 182 | -35.2% |
| DualStack | 2 | 195 | 115 | -41% |
| PSM Light | 1 | 5 | 5 | 0%  |

Concurrency in LB suite on master is slightly reduced due to a long bootstrap generator test that runs at the end.

---

### Test Runs
* Security: [Logs](https://source.cloud.google.com/results/invocations/861c691e-c1c6-4532-9532-374da59749ff)
* LB: [Logs](https://source.cloud.google.com/results/invocations/19f55780-19f5-4f50-8d94-f30ca52dca5f)
* DualStack: [Logs](https://source.cloud.google.com/results/invocations/0ca1e9a6-a0d5-4157-8f02-45c9b6a788ed)
* PSM Light (unaffected): [Logs](https://source.cloud.google.com/results/invocations/b2be5f78-e641-4e9a-a257-c758776a44ed)

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -519,12 +519,36 @@ psm::run() {
 psm::run::test_suite() {
   local test_suite="${1:?${FUNCNAME[0]} missing the test suite argument}"
   cd "${TEST_DRIVER_FULL_DIR}"
+
+  # Export variables needed by subshells
+  export TEST_XML_OUTPUT_DIR
+  export TEST_DRIVER_FLAGFILE
+  export KUBE_CONTEXT
+  export TESTING_VERSION
+  export CLIENT_IMAGE_NAME
+  export GIT_COMMIT
+  export SERVER_IMAGE_USE_CANONICAL
+  export SERVER_IMAGE_NAME
+  export SECONDARY_KUBE_CONTEXT
+  export GRPC_LANGUAGE
+  export PSM_EXTRA_FLAGS
+  export VIRTUAL_ENV
+
+  # Export functions needed by subshells
+  export -f psm::tools::log
+  export -f psm::tools::run_verbose
+  export -f psm::run::finalize_test_flags
+  export -f "psm::${test_suite}::run_test"
+  export -f psm::run::test
+
   local failed_tests=0
-  for test_name in "${TESTS[@]}"; do
-    psm::run::test "${test_suite}" "${test_name}" || (( ++failed_tests ))
-    psm::tools::log "Finished ${test_suite} suite test: ${test_name}"
-    echo
-  done
+  local jobs="${PSM_PARALLEL_JOBS:-2}"
+
+  psm::tools::log "Running ${test_suite} suite tests in parallel with ${jobs} jobs"
+  # We use --line-buffer to see output in real-time, preventing half-lines from mixing.
+  # We use --shell bash to ensure it runs in bash.
+  parallel --shell bash --line-buffer --jobs "${jobs}" psm::run::test "${test_suite}" ::: "${TESTS[@]}" || failed_tests=$?
+
   psm::tools::log "Failed test suites: ${failed_tests}"
 }
 
@@ -575,8 +599,11 @@ psm::run::test() {
   fi
 
   psm::tools::log "Running ${test_suite} suite test: ${test_name}" |& tee "${test_log}"
-  # Must be the last line.
-  "psm::${test_suite}::run_test" "${test_name}" |& tee -a "${test_log}"
+  local exit_code=0
+  "psm::${test_suite}::run_test" "${test_name}" |& tee -a "${test_log}" || exit_code=$?
+  psm::tools::log "Finished ${test_suite} suite test: ${test_name}"
+  echo
+  return ${exit_code}
 }
 
 #######################################
@@ -1278,7 +1305,8 @@ kokoro_install_dependencies() {
   sudo DEBIAN_FRONTEND=noninteractive apt-get -qq install --auto-remove \
     "python${PYTHON_VERSION}-venv" \
     google-cloud-sdk-gke-gcloud-auth-plugin \
-    kubectl
+    kubectl \
+    parallel
   sudo rm -rf /var/lib/apt/lists
 }
 
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -575,6 +575,7 @@ psm::run::test_suite() {
 #   Test xUnit report to ${TEST_XML_OUTPUT_DIR}/${test_name}/sponge_log.xml
 #######################################
 psm::run::test() {
+  set -eo pipefail
   # Test driver usage: https://github.com/grpc/psm-interop#basic-usage
   local test_suite="${1:?${FUNCNAME[0]} missing the test suite argument}"
   local test_name="${2:?${FUNCNAME[0]} missing the test name argument}"
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -520,7 +520,7 @@ psm::run::test_suite() {
   local test_suite="${1:?${FUNCNAME[0]} missing the test suite argument}"
   cd "${TEST_DRIVER_FULL_DIR}"
 
-  # Export variables needed by subshells
+  # Export variables needed by subshells.
   export TEST_XML_OUTPUT_DIR
   export TEST_DRIVER_FLAGFILE
   export KUBE_CONTEXT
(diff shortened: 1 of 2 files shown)


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -519,12 +519,41 @@ psm::run() {
...
+       export -f "$func"
+   done
+
   local failed_tests=0
-  for test_name in "${TESTS[@]}"; do
-    psm::run::test "${test_suite}" "${test_name}" || (( ++failed_tests ))
-    psm::tools::log "Finished ${test_suite} suite test: ${test_name}"
-    echo
-  done
+  # TODO(b/526886100): Make the parallelism configurable using a function
```
pawbhard: Create a separate task for it

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -575,8 +604,11 @@ psm::run::test() {
   fi
 
   psm::tools::log "Running ${test_suite} suite test: ${test_name}" |& tee "${test_log}"
-  # Must be the last line.
-  "psm::${test_suite}::run_test" "${test_name}" |& tee -a "${test_log}"
+  local exit_code=0
```
pawbhard, line 608: Can we have a verify run with one of the test failing so we can check if that is handled correctly by the logic

pawbhard (commented):
LGTM, Need One validation for failure case.

@sergiitk , DO you have any concerns as you were trying this earlier, else we can merge.
We will run for a week for these test suites and then enable for all with configurable parallelism

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: I injected an error to cause the dualstack test to fail and verified that the failure is reported correctly, however the the counter in the final log is 0
> Failed test suites: 0

I'm looking into this.
Run : https://source.cloud.google.com/results/invocations/922f81cd-5083-45b8-8784-8b5716559e37

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: Filed b/535109852.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: The issue was caused because `-eo piepefail` wasn't getting propagated to the subshells launched by parallel. I've fixed it now: https://source.cloud.google.com/results/invocations/0fd0cdb4-a860-40ca-b37f-2ed94eb2cef6

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -519,12 +519,41 @@ psm::run() {
...
+  case "${test_suite}" in
+    security | lb | dualstack)
+      jobs=2
+      ;;
+  esac
+
+  psm::tools::log "Running ${test_suite} suite tests in parallel with ${jobs} jobs"
+  # We use --line-buffer to see output in real-time, preventing half-lines from
+  # mixing.
+  parallel --line-buffer --jobs "${jobs}" psm::run::test "${test_suite}" ::: "${TESTS[@]}" || failed_tests=$?
```
pawbhard, line 555: earlier we had "(( ++failed_tests ))"  here we are overwritting

pawbhard (commented):
thanks for testing and fixing failure logic.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: Incrementing the same variable from concurrently running subshells will cause data races. Luckily the return value of of GNU parallel is the number of failed jobs, so we can use that directly.

pawbhard: ACK

pawbhard (approved):

.kokoro/psm_interop_kokoro_lib.sh
sergiitk, line 555: Nice move switching to `parallel`! One subtle detail around CI status handling: GNU `parallel` returns `1..100` for failed jobs (capped at `101`), but returns `254`/`255` if `parallel` itself hits a syntax error, OOM, or signal crash.

Right now, since `psm::tools::log` is the last command in `psm::run::test_suite()`, the function always returns `0`. If `parallel` crashes (`255`) before spawning the Python runners, Kokoro will get exit code `0` with no `sponge_log.xml` files generated, which could mask an infrastructure crash as `PASSED (0 tests)`.

What do you think about explicitly returning non-zero if `parallel` hits a runner/tool crash (`status > 100`) while keeping `0` for standard test failures so Kokoro can parse `sponge_log.xml`?

```sh
# If parallel itself crashed (>100), fail the script immediately so Kokoro catches the infra error
if (( status > 101 )); then
  psm::tools::log "Error: GNU parallel crashed or exited abnormally with status ${status}"
  return ${status}
fi
if (( status == 101 )); then
  psm::tools::log "Failed test cases in ${test_suite}: >100"
else
  psm::tools::log "Failed test cases in ${test_suite}: ${status}"
fi

```

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -575,8 +605,11 @@ psm::run::test() {
   fi
 
   psm::tools::log "Running ${test_suite} suite test: ${test_name}" |& tee "${test_log}"
-  # Must be the last line.
-  "psm::${test_suite}::run_test" "${test_name}" |& tee -a "${test_log}"
+  local exit_code=0
```
sergiitk, line 608: One edge case to watch out for: with `-o pipefail`, the initial logging command above is unprotected. If `tee` encounters an error (like a transient disk full or broken pipe), `pipefail` will cause that line to return non-zero and `set -e` will instantly abort the subshell before `run_test` even executes.

To make sure a logging/tee glitch on startup doesn't kill the subshell before our actual test runs, what do you think about protecting that initial `tee` pipeline with `|| true` (or capturing its status)?

```sh
psm::tools::log "Running ${test_suite} suite test: ${test_name}" |& tee "${test_log}" || true
local exit_code=0
...
```

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -519,12 +519,41 @@ psm::run() {
 psm::run::test_suite() {
   local test_suite="${1:?${FUNCNAME[0]} missing the test suite argument}"
   cd "${TEST_DRIVER_FULL_DIR}"
+
+  # Export variables needed by subshells.
```
sergiitk, line 523: I like that it's clear what's actually used now. A few things:

- Let's `export PS4` too.
- Perhaps, order alphabetically?

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -519,12 +519,41 @@ psm::run() {
...
+  # parameter.
+  local jobs=1
+  case "${test_suite}" in
+    security | lb | dualstack)
+      jobs=2
+      ;;
+  esac
+
+  psm::tools::log "Running ${test_suite} suite tests in parallel with ${jobs} jobs"
+  # We use --line-buffer to see output in real-time, preventing half-lines from
```
sergiitk, line 553: Not sure `--line-buffer` vs `--group` (which is the default). Things py like stack traces will look bad. I guess we can rely on per-test logs...

sergiitk (commented):
This is great! :shipit:  
Happy about the run time.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: Done.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: `--line-buffer` prints logs from each job in real time while `--group` buffers the logs until the job is complete and prints the logs together. This means that there will be no logs for the time a job is running in the background making it harder to debug in real time.

The staging tests also interleave log lines from different test cases to stdout. The individual test case logs can be seen in the test artifacts.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: Agree, changed as suggested.

.kokoro/psm_interop_kokoro_lib.sh
arjan-bal: Done. This seems to be an existing edge case, earlier `-e` was being set in another file, not it's set again in the subshell.

arjan-bal (commented):
Raised https://github.com/grpc/psm-interop/pull/266 to address the comments.

PR #268, opened 2026-07-21
Update the version logic and tagging

pawbhard:
Update the logic to determine if we need to tag the image
Also added a check to retag version if not present for pre-built image

Test for bash logic 
```
bpawan@bpawan:~/workspace/code/psm-interop$ for val in kokoro kokoro-github-subscriber anonymous user123; do                                                                                                                                                                                              
      KOKORO_BUILD_INITIATOR="$val"                                                                                                                                                                                                                               
      if [[ "${KOKORO_BUILD_INITIATOR:-anonymous}" != kokoro* ]]; then                                                                                                                                                                                            
        echo "$val: DEV"                                                                                                                                                                                                                                          
      else                                                                                                                                                                                                                                                        
        echo "$val: OFFICIAL"                                                                                                                                                                                                                                     
      fi                                                                                                                                                                                                                                                          
    done   
kokoro: OFFICIAL
kokoro-github-subscriber: OFFICIAL
anonymous: DEV
user123: DEV
bpawan@bpawan:~/workspace/code/psm-interop$ 
```

### DIFF
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -822,15 +822,13 @@ psm::build::docker_images_if_needed() {
     } |& tee -a "${BUILD_LOGS_ROOT}/build-docker.log"
   else
     psm::tools::log "Skipping ${GRPC_LANGUAGE} test app build"
-    # Image exists. Check if version is tagged, if not tag it.
+    # Image exists but version is not correctly tagged
     if is_version_branch "${TESTING_VERSION}"; then
-      if [[ "${client_tags}" != *"${TESTING_VERSION}"* ]]; then
-        psm::tools::log "Adding version tag ${TESTING_VERSION} to existing client image ${CLIENT_IMAGE_NAME}:${GIT_COMMIT}"
-        psm::tools::run_verbose gcloud -q container images add-tag "${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" "${CLIENT_IMAGE_NAME}:${TESTING_VERSION}"
-      fi
-      if [[ -z "${SERVER_IMAGE_USE_CANONICAL}" && "${server_tags}" != *"${TESTING_VERSION}"* ]]; then
-        psm::tools::log "Adding version tag ${TESTING_VERSION} to existing server image ${SERVER_IMAGE_NAME}:${GIT_COMMIT}"
-        psm::tools::run_verbose gcloud -q container images add-tag "${SERVER_IMAGE_NAME}:${GIT_COMMIT}" "${SERVER_IMAGE_NAME}:${TESTING_VERSION}"
+      client_tags_raw="$(gcloud_gcr_list_raw_image_tags "${CLIENT_IMAGE_NAME}" "${GIT_COMMIT}")"
+      gcloud_gcr_add_version_tag_if_missing "${CLIENT_IMAGE_NAME}" "${client_tags_raw}" "${TESTING_VERSION}"
+      if [[ -z "${SERVER_IMAGE_USE_CANONICAL}" ]]; then
+        server_tags_raw="$(gcloud_gcr_list_raw_image_tags "${SERVER_IMAGE_NAME}" "${GIT_COMMIT}")"
+        gcloud_gcr_add_version_tag_if_missing "${SERVER_IMAGE_NAME}" "${server_tags_raw}" "${TESTING_VERSION}"
       fi
     fi
   fi
@@ -1085,6 +1083,40 @@ is_version_branch() {
 #######################################
 gcloud_gcr_list_image_tags() {
   gcloud container images list-tags --format="table[box](tags,digest,timestamp.date())" --filter="tags:$2" "$1"
+}
+
+#######################################
+# List GCR image tags matching given tag name in raw format.
+# Arguments:
+#   Image name
+#   Tag name
+# Outputs:
+#   Writes comma-delimited list of tags to stdout.
+#   If no tags found, the output is an empty string.
+#######################################
+gcloud_gcr_list_raw_image_tags() {
+  gcloud container images list-tags --format="value(tags)" --filter="tags:$2" "$1"
+}
+
+#######################################
+# Adds a version tag to an existing GCR image if missing.
+# Globals:
+#   GIT_COMMIT: SHA-1 of git commit being built
+# Arguments:
+#   Image name
+#   Comma-separated list of existing tags
+#   Tag to add if missing
+# Outputs:
+#   Writes logs to stdout
+#######################################
+gcloud_gcr_add_version_tag_if_missing() {
+  local image_name="$1"
+  local existing_tags="$2"
+  local to_tag="$3"
+  if [[ ! "${existing_tags}" =~ (^|,)${to_tag}(,|$) ]]; then
+    psm::tools::log "Adding version tag ${to_tag} to existing image ${image_name}:${GIT_COMMIT}"
+    psm::tools::run_verbose gcloud -q container images add-tag "${image_name}:${GIT_COMMIT}" "${image_name}:${to_tag}"
+  fi
 }
 
 #######################################
--- a/.kokoro/psm_interop_kokoro_lib.sh
+++ b/.kokoro/psm_interop_kokoro_lib.sh
@@ -822,8 +822,7 @@ psm::build::docker_images_if_needed() {
     } |& tee -a "${BUILD_LOGS_ROOT}/build-docker.log"
   else
     psm::tools::log "Skipping ${GRPC_LANGUAGE} test app build"
-    # Image exists but version is not correctly tag
-    # Tag the version
+    # Image exists. Check if version is tagged, if not tag it.
     if is_version_branch "${TESTING_VERSION}"; then
       if [[ "${client_tags}" != *"${TESTING_VERSION}"* ]]; then
         psm::tools::log "Adding version tag ${TESTING_VERSION} to existing client image ${CLIENT_IMAGE_NAME}:${GIT_COMMIT}"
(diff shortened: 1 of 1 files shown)


.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -1351,7 +1363,7 @@ kokoro_get_testing_version() {
     # Allows to override the testing version, and force tagging the built
     # images, if necessary.
     readonly TESTING_VERSION="${PSM_FORCE_TESTING_VERSION}"
-  elif [[ "${KOKORO_BUILD_INITIATOR:-anonymous}" != "kokoro" ]]; then
+  elif [[ "${KOKORO_BUILD_INITIATOR:-anonymous}" != kokoro* ]]; then
```
arjan-bal, line 1395: Can we use exact names instead of doing a prefix match?

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -822,6 +822,18 @@ psm::build::docker_images_if_needed() {
     } |& tee -a "${BUILD_LOGS_ROOT}/build-docker.log"
   else
     psm::tools::log "Skipping ${GRPC_LANGUAGE} test app build"
+    # Image exists but version is not correctly tag
+    # Tag the version
```
arjan-bal: Comments should be complete sentences that end in a period. The grammar seems to be off here.

.kokoro/psm_interop_kokoro_lib.sh
sergiitk: The prefix should be fine, there are too many names for kokoro to track here. I have the same logic here: https://github.com/grpc/grpc/blob/21698ef14a11d7939633948ae5ffe42700e7db8a/tools/run_tests/python_utils/utils_common.py#L32-L35.

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Done

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: prefer keeping prefix

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -822,6 +822,17 @@ psm::build::docker_images_if_needed() {
     } |& tee -a "${BUILD_LOGS_ROOT}/build-docker.log"
   else
     psm::tools::log "Skipping ${GRPC_LANGUAGE} test app build"
+    # Image exists. Check if version is tagged, if not tag it.
```
sergiitk: typo

```suggestion
    # Image exists but version is not correctly tagged
```

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -822,6 +822,17 @@ psm::build::docker_images_if_needed() {
...
+    if is_version_branch "${TESTING_VERSION}"; then
+      if [[ "${client_tags}" != *"${TESTING_VERSION}"* ]]; then
```
sergiitk: Bare substring matching against `gcloud_gcr_list_image_tags` couples automated CI flow to ASCII display formatting and introduces a pipeline bug.

If an image was previously tagged in a dev build (e.g. `dev-master` or `dev-v1.82.x`), a later official build for `master` or `v1.82.x` will trigger a false positive and silently skip release image tagging.

To decouple programmatic CI logic from visual log tables and prevent tag collisions, consider adding a non-table helper function:

```bash
gcloud_gcr_list_raw_image_tags() {
  gcloud container images list-tags --format="value(tags)" --filter="tags:$2" "$1"
}
```

Because `--format="value(tags)"` produces a comma-delimited string (e.g. `96b6849...,v1.82.x`), tag presence can be cleanly checked using an anchored regex:

```bash
if [[ ! "${client_tags}" =~ (^|,)${TESTING_VERSION}(,|$) ]]; then
```

Same applies to `server_tags` below.

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -822,6 +822,17 @@ psm::build::docker_images_if_needed() {
...
+        psm::tools::log "Adding version tag ${TESTING_VERSION} to existing client image ${CLIENT_IMAGE_NAME}:${GIT_COMMIT}"
+        psm::tools::run_verbose gcloud -q container images add-tag "${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" "${CLIENT_IMAGE_NAME}:${TESTING_VERSION}"
+      fi
+      if [[ -z "${SERVER_IMAGE_USE_CANONICAL}" && "${server_tags}" != *"${TESTING_VERSION}"* ]]; then
+        psm::tools::log "Adding version tag ${TESTING_VERSION} to existing server image ${SERVER_IMAGE_NAME}:${GIT_COMMIT}"
+        psm::tools::run_verbose gcloud -q container images add-tag "${SERVER_IMAGE_NAME}:${GIT_COMMIT}" "${SERVER_IMAGE_NAME}:${TESTING_VERSION}"
```
sergiitk: nit: 

The conditional checks and remote image tagging logic are currently duplicated for client and server targets. To reduce structural duplication and align with the design of tag_and_push_docker_image, consider extracting a shared helper function for remote tagging in this library:

```bash
gcloud_gcr_add_version_tag_if_missing() {
  local image_name="$1"
  local existing_tags="$2"
  local to_tag="$3"
  if [[ "${existing_tags}" != *"${to_tag}"* ]]; then
    psm::tools::log "Adding version tag ${to_tag} to existing image ${image_name}:${GIT_COMMIT}"
    psm::tools::run_verbose gcloud -q container images add-tag "${image_name}:${GIT_COMMIT}" "${image_name}:${to_tag}"
  fi
}
```

Then use `psm gcloud_gcr_add_version_tag_if_missing "${CLIENT_IMAGE_NAME}" "${client_tags}" "${TESTING_VERSION}"` and a single conditional call for the server image to keep the main flow concise.

sergiitk: Update: see my other comment on the clean tag load and the logic for `if`

sergiitk (changes requested):

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Addressed

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Done

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Done

pawbhard (commented):
Addressed comments , @sergiitk PTAL

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -822,6 +822,15 @@ psm::build::docker_images_if_needed() {
     } |& tee -a "${BUILD_LOGS_ROOT}/build-docker.log"
   else
     psm::tools::log "Skipping ${GRPC_LANGUAGE} test app build"
+    # Image exists but version is not correctly tagged
```
sergiitk: nit: another iteration on the comment

```suggestion
    # Image exists; ensure it is tagged with the testing version if on a version branch
```

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -1076,6 +1085,40 @@ gcloud_gcr_list_image_tags() {
...
+#######################################
+gcloud_gcr_add_version_tag_if_missing() {
+  local image_name="$1"
+  local existing_tags="$2"
+  local to_tag="$3"
+  if [[ ! "${existing_tags}" =~ (^|,)${to_tag}(,|$) ]]; then
+    psm::tools::log "Adding version tag ${to_tag} to existing image ${image_name}:${GIT_COMMIT}"
+    psm::tools::run_verbose gcloud -q container images add-tag "${image_name}:${GIT_COMMIT}" "${image_name}:${to_tag}"
+  fi
+}
```
sergiitk, line 1118: improvement: Moving `gcloud_gcr_list_raw_image_tags` inside this helper simplifies caller logic and keeps parameter signatures consistent with `tag_and_push_docker_image`.

By explicitly accepting `from_tag` as `$2` and `to_tag` as `$3`, the helper decouples from global `GIT_COMMIT` and manages tag fetching internally without polluting caller scopes with temporary variables. Applying parameter expansion checks (`${1:?}`) fails fast if required arguments are omitted.

```suggestion
gcloud_gcr_add_version_tag_if_missing() {
  local image_name="${1:?image_name is required}"
  local from_tag="${2:?from_tag is required}"
  local to_tag="${3:?to_tag is required}"
  local existing_tags
  existing_tags="$(gcloud_gcr_list_raw_image_tags "${image_name}" "${from_tag}")"
  if [[ ! "${existing_tags}" =~ (^|,)${to_tag}(,|$) ]]; then
    psm::tools::log "Adding version tag ${to_tag} to existing image ${image_name}:${from_tag}"
    psm::tools::run_verbose gcloud -q container images add-tag "${image_name}:${from_tag}" "${image_name}:${to_tag}"
  fi
}
```

In `docker_images_if_needed`, invocations for client and server images simplify to calling `gcloud_gcr_add_version_tag_if_missing "${IMAGE_NAME}" "${GIT_COMMIT}" "${TESTING_VERSION}"` directly without pre-fetching tag lists.

sergiitk (approved):

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Done

.kokoro/psm_interop_kokoro_lib.sh
pawbhard: Done

PR #269, opened 2026-07-22
[PSM Interop] Add cleanup for orphaned cloud run NEGs and backend services.

Pranjali-2501:
This PR add support for cleaning up leaked/orphaned Google Cloud Run resources within the cleanup.py script. It adds a new "cloudrun" mode that scans for and deletes expired Cloud Run services, serverless NEGs (Network Endpoint Groups), and regional backend services created during interop tests.

### DIFF
--- a/bin/cleanup/cleanup.py
+++ b/bin/cleanup/cleanup.py
@@ -121,8 +121,8 @@ CLIENT_PREFIXES = flags.DEFINE_list(
 MODE = flags.DEFINE_enum(
     "mode",
     default="td",
-    enum_values=["k8s", "td", "td_no_legacy"],
-    help="Mode: Kubernetes or Traffic Director",
+    enum_values=["k8s", "td", "td_no_legacy", "cloudrun"],
+    help="Mode: Kubernetes, Traffic Director, or Cloud Run",
 )
 SECONDARY = flags.DEFINE_bool(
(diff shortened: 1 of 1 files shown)


bin/cleanup/cleanup.py
```diff
@@ -791,6 +791,57 @@ def find_and_remove_leaked_td_resources(
         )
 
 
+def find_and_remove_leaked_cloudrun_resources(dry_run, project, region):
```
sreenithi, line 794: I think it would be better to have a cleanup logic similar to other functions in this file like `find_and_remove_leaked_k8s_resources`, `find_and_remove_leaked_td_resources` where we create the respective classes and call its cleanup() functions.

So in this case, you will have to create separate functions similar to `cleanup_client`, `cleanup_server` that creates the `CloudRunServerRunner` and `CloudRunClientRunner` objects and calls their respective `cleanup()` functions.

Similarly for the NEGs, you can possibly modify the existing `cleanup_td_for_gke` function to create a `CloudRunMeshManager` class and call its `cleanup()` function.

This way we can stay consistent with the remaining cleanup jobs in the file, and secondly if in future, if we had to introduce any new resource for the tests, the respective cleanup() functions will automatically handle it and people will not have to specifically update the cleanup jobs again.

PR #271, opened 2026-07-27
Python 3.11 - Using uv and bump dependencies for 3.11

asheshvidyut:
# Description

This PR installs `uv` and then installs `Python 3.11` using `uv`.

Virtual env is also created using `uv` with `Python 3.11`. 

# Testing

* PSM Security: [#1751 Logs](https://source.cloud.google.com/results/invocations/e3e8a0e4-4a3d-4a8b-a799-79d48e4288a8)
* XDS K8s LB: [#1667 Logs](https://source.cloud.google.com/results/invocations/c5ee15af-6eb4-4e99-b9b5-64cf737c676a)
* Dual Stack: [#555 Logs](https://source.cloud.google.com/results/invocations/f50a123a-8bd3-48cb-a0f4-2b3e7e531108)
* PSM Light: [#451 Logs](https://source.cloud.google.com/results/invocations/cdecbc68-3418-47b7-9fdd-36158d33b937)


framework/helpers/retryers.py
```diff
@@ -235,9 +234,7 @@ def log_it(retry_state):
             verb, value = "raised", "%s: %s" % (type(ex).__name__, ex)
 
             if exc_info:
-                local_exc_info = tenacity_compat.get_exc_info_from_future(
-                    retry_state.outcome
-                )
+                local_exc_info = ex
```
asheshvidyut, line 237: This function `get_exc_info_from_future` is removed from [tenacity](https://github.com/search?q=repo%3Ajd%2Ftenacity%20get_exc_info_from_future&type=code)

framework/test_app/runners/k8s/k8s_base_runner.py
```diff
@@ -1002,7 +1002,7 @@ def _wait_pod_started(self, name, **kwargs) -> k8s.V1Pod:
         self.k8s_namespace.wait_for_pod_started(name, **kwargs)
         pod = self.k8s_namespace.get_pod(name)
 
-        pod_ips = pod.status.pod_i_ps or pod.status.pod_ip
+        pod_ips = pod.status.pod_ips or pod.status.pod_ip
```
asheshvidyut, line 1005: As per latest docs its `pod_ips` and `pod_ip`

[kubernetes/docs/V1PodStatus.md](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1PodStatus.md)

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -1289,8 +1289,7 @@ test_driver_install() {
 kokoro_print_version() {
   psm::tools::log "Kokoro Ubuntu version:"
   run_ignore_exit_code lsb_release -a
-  run_ignore_exit_code "python${PYTHON_VERSION}" --version
-  run_ignore_exit_code "python${PYTHON_VERSION}" -m pip --version
```
sreenithi, line 1293: let's continue to keep this here to log what Python version is coming from the system. 

However add another log line above to say this is System Python version and may be different from the actual running version

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -1360,11 +1359,15 @@ kokoro_install_dependencies() {
...
-    "python${PYTHON_VERSION}-venv" \
     google-cloud-sdk-gke-gcloud-auth-plugin \
     kubectl \
     parallel
   sudo rm -rf /var/lib/apt/lists
+  sudo python3 -m pip install uv
+  uv python install "${PYTHON_VERSION}"
+  psm::tools::log "uv version:"
+  uv --version
+  uv run --python "${PYTHON_VERSION}" python --version
```
sreenithi, line 1375: Let's add the same log lines in `test_driver_pip_install()` also above. Reason being, the output of this `kokoro_install_dependencies()` function seems to be getting routed to a different log file ([Ref](https://github.com/asheshvidyut/psm-interop/blob/646f4d038d6835000f5f35e4b17b374a113f3722/.kokoro/psm_interop_kokoro_lib.sh#L1468)) and hence we wouldn't be able to see this directly in the logs.

requirements.txt
```diff
@@ -1,25 +1,26 @@
...
-six~=1.13
-tenacity~=6.2
-typing-extensions~=4.10
-packaging~=23.1
-Pygments~=2.9
-python-dateutil~=2.8
-protobuf~=6.31
-requests~=2.31.0
-xds-protos~=1.74
+google-auth==2.43.0
```
sreenithi: I believe the main dependency we wanted to update was httplib2 as per b/529575020

But since we had the issue as mentioned in b/492457764#comment16, can you try if the issue is resolved if we update this to something more recent now.

.kokoro/psm_interop_kokoro_lib.sh
```diff
@@ -1289,6 +1292,7 @@ test_driver_install() {
 kokoro_print_version() {
   psm::tools::log "Kokoro Ubuntu version:"
   run_ignore_exit_code lsb_release -a
+  psm::tools::log "This is System Python version and may be different from actual running version"
```
sreenithi: nit:
```suggestion
  psm::tools::log "This is the System Python version and may be different from the actual running version"
```

.kokoro/psm_interop_kokoro_lib.sh
sreenithi: Looks like there is a problem with this now. The log lines say:
```
+ [09:56:42 UTC]	 This is System Python version and may be different from actual running version
/tmp/tmp.Z8BbXKVkuU/psm_interop_kokoro_lib.sh: line 1028: python3.11: command not found
+ [09:56:42 UTC]	 Cmd: 'python3.11 --version', exit code: 127
/tmp/tmp.Z8BbXKVkuU/psm_interop_kokoro_lib.sh: line 1028: python3.11: command not found
+ [09:56:42 UTC]	 Cmd: 'python3.11 -m pip --version', exit code: 127
```

Sorry to make you do a lot of effort for this thing. Can you just try saying `python3` here instead of `"python${PYTHON_VERSION}"` and run one of the tests. In case it still doesn't work, you can remove it.

sreenithi (approved):
LGTM, just some last comments

.kokoro/psm_interop_kokoro_lib.sh
asheshvidyut: Sure

asheshvidyut: No issues. This is fixed now.

bin/freeze.sh
```diff
@@ -16,13 +16,5 @@
...
-readonly VENV_NAME
-
-python3 -m virtualenv "${VENV_NAME}"
-
-"${VENV_NAME}"/bin/pip install -r requirements.txt
-"${VENV_NAME}"/bin/pip freeze --require-virtualenv --local -r requirements.txt \
-  > requirements.lock
-
-rm -rf "${VENV_NAME}"
+# Make sure uv is installed: curl -LsSf https://astral.sh/uv/install.sh | sh
```
asheshvidyut: Should we install `uv` here?

@sreenithi

sreenithi: Sure, good catch. But let's use pip to install not this shell script.

framework/helpers/retryers.py
pawbhard: What will the impact of this change ?
Also check, are we using tenacity lib for some other work ? DO we need to update it or remove?

.pylintrc
```diff
@@ -104,3 +104,24 @@ disable=
 	unsubscriptable-object,
 	# NOTE(sergiitk): yapf compatibility, ref #25071
 	bad-continuation,
+	# TODO: Clean up the following suppressions and fix them in code later
```
pawbhard: create a tak for this on yourself add ldap and bug to it

pawbhard (approved):
LGTM, Just check once b/529575020, this is the bug because of which upgrade is happening, check if we are updating decency mentioned in bug (If needed check with reporter).

asheshvidyut:
> LGTM, Just check once b/529575020, this is the bug because of which upgrade is happening, check if we are updating decency mentioned in bug (If needed check with reporter).

Yes we are updating `google-auth-httplib2` to the latest version [0.4.0](https://pypi.org/project/google-auth-httplib2/0.4.0/)

.pylintrc
asheshvidyut: Done

asheshvidyut:
Regarding the `tenacity_compat.get_exc_info_from_future`

This is the PR which drops it - https://github.com/jd/tenacity/pull/304/changes

source code

```python
def get_exc_info_from_future(future):
    """
    Get an exc_info value from a Future.

    Given a a Future instance, retrieve an exc_info value suitable for passing
    in as the exc_info parameter to logging.Logger.log() and related methods.

    On Python 2, this will be a (type, value, traceback) triple.
    On Python 3, this will be an exception instance (with embedded traceback).

    If there was no exception, None is returned on both versions of Python.
    """
    if six.PY3:
        return future.exception()
    else:
        ex, tb = future.exception_info()
        if ex is None:
            return None
        return type(ex), ex, tb
```

This is called with `retry_state.outcome` which is already done in the line above. Hence I am just using `local_exc_info = ex`

Hence no impact as such.

framework/helpers/retryers.py
asheshvidyut: Commented on PR. https://github.com/grpc/psm-interop/pull/271#issuecomment-5139941958

